PR #22, opened 2020-03-25
Create alias cmake target

LesleyLai:


charles-lunarg:
I will merge this, but I would like a brief explanation of what/why alias targets are needed.

LesleyLai:
It is kind of a CMake best practice thing. If you have a typo for targets with a namespace, it will be catched in the configuration time instead of link time.

PR #23, opened 2020-03-26
Migrate to Catch2 & CMake file structural change

LesleyLai:
Sorry about this big commit.

- Use Catch to do nested testing
- Make vulkan include path public so the tests and examples can find it
- Make compiler warnings an interface target so the tests and example can also use it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 *.vscode
 *.vs
+*.idea
+cmake-build-*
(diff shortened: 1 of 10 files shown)


charles-lunarg:
Nothing wrong with doing some much needed repo maintenance.
Thank you VERY MUCH for doing this! Especially the cmake, wouldn't of thought to make compiler options its own interface target.
Lots of little things that should of been done, but because they 'worked' I didn't even think about it.
And now that I see how you structured the catch2 tests, adding more will be straightforward.
Going to do a second read through to make sure I understand everything but looks great so far.

charles-lunarg:
Nothing wrong with doing some much needed repo maintenance.
Thank you VERY MUCH for doing this! Especially the cmake, wouldn't of thought to make compiler options its own interface target.
Lots of little things that should of been done, but because they 'worked' I didn't even think about it.
And now that I see how you structured the catch2 tests, adding more will be straightforward.
Going to do a second read through to make sure I understand everything but looks great so far.

Issue #24, opened 2020-04-12
Refactor the `Expected` type to a boost outcome like `Result` type


LesleyLai:
The current implementation of `Expected` in the header is a general-purpose type that mimics the proposed `std::expected`. I propose to change the `Expected<T, E>` to a `Result<T>` type which uses `std::error_code` to represent the error. There are several advantages for this design:
1. We can add Vulkan specific APIs to our `Result` class
2. More ergonomic error reporting with `<system_error>`
3. A type-erased error-code type makes chaining errors easier (especially if we want to pursue monadic extension later)

The downside:
1. Depending on `std::error_code` means pulling in the `<system_error>` header

charles-lunarg:
Thinking back to the initial versions of the expected type, the issue I came across was handling custom vk-bootstrap errors alongside vulkan specific VkResult errors that came from vulkan functions call inside. I believe the issue was that I didn't have type-erasure, so I had to make a custom type which held VkResult's. 
Error handling has been something I would like to see revisited, as the current approach is extremely simplistic and really only handles the 'oops I died' case with little ability to know why.

So this would be a welcome PR and should make for a much more robust library.
And since system_error is C++11, there is no issue in adding it. While I am trying to reduce unnecessary headers, `<system_error>` would make many things simpler with it than without it.

PR #31, opened 2020-05-17
Error handling with std::error_code

LesleyLai:
Address #24.

**Warning**: API break

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -11,7 +11,7 @@ Simply create a builder variable and call the `build()` member function.
 vkb::InstanceBuilder instance_builder;
 auto instance_builder_return = instance_builder.build();
 ```
-Because creating an instance may fail, the builder returns an 'Expected' type. This contains either a valid `vkb::Instance` struct, which includes a `VkInstance` handle, or contains an `vkb::InstanceError`.
+Because creating an instance may fail, the builder returns an 'Result' type. This contains either a valid `vkb::Instance` struct, which includes a `VkInstance` handle, or contains an `vkb::InstanceError`.
 ```cpp
 if (!instance_builder_return) {
     printf("Failed to create Vulkan instance. Cause %s\n", 
@@ -19,7 +19,7 @@ if (!instance_builder_return) {
     return -1;
 } 
 ```
-Once any possible errors have been dealt with, we can pull the `vkb::Instance` struct out of the `Expected`. 
+Once any possible errors have been dealt with, we can pull the `vkb::Instance` struct out of the `Result`. 
 ```cpp
 vkb::Instance vkb_instance = instance_builder_return.value();
 ```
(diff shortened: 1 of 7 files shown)


src/VkBootstrap.h
```diff
@@ -131,6 +132,13 @@ enum class SwapchainError {
 	failed_get_swapchain_images,
 	failed_create_swapchain_image_views,
 };
+
+std::error_code make_error_code (InstanceError instance_error);
```
charles-lunarg, line 136: This should be in the cpp file, since its purely an implementation helper.

docs/getting_started.md
```diff
@@ -11,15 +11,15 @@ Simply create a builder variable and call the `build()` member function.
 vkb::InstanceBuilder instance_builder;
 auto instance_builder_return = instance_builder.build();
 ```
-Because creating an instance may fail, the builder returns an 'Expected' type. This contains either a valid `vkb::Instance` struct, which includes a `VkInstance` handle, or contains an `vkb::InstanceError`.
+Because creating an instance may fail, the builder returns an 'Result' type. This contains either a valid `vkb::Instance` struct, which includes a `VkInstance` handle, or contains an `vkb::InstanceError`.
 ```cpp
 if (!instance_builder_return) {
     printf("Failed to create Vulkan instance. Cause %s\n", 
         vkb::to_string(instance_builder_return.error().type));
```
charles-lunarg, line 18: Should be updated, not a must fix.
Likely the readme.md also needs an looking over, but that can come later too.

src/VkBootstrap.h
LesleyLai: Making this in the cpp files will break the tests.

cdgiessen: Ah, okay, but can we put it in a detail namespace or something? I'm hoping to keep only whats necessary exposed directly in the vkb namespace.

charles-lunarg: Ah, okay, but can we put it in a detail namespace or something? I'm hoping to keep only whats necessary exposed directly in the vkb namespace.

LesleyLai: That is an option. Though if the client-side cannot find `make_error_code` through ADL, it is not technically correct usage of `std::error_code`.

charles-lunarg: If you think this needs to stay in the vkb namespace, I am not against it. I just feel is good to try to limit the api as much as possible.

Plus, if there is a reason to keep it in the vkb namespace, you don't have to make any changes, meaning I can merge his then be able to start working on documenting the change.

LesleyLai: Yeah, I think it provides some value the user can convert your enums into `std::error_code` themselves. Plus putting in another namespace would complicate the implementation (as `std::error_code` cannot find them automatically.)

charles-lunarg: Sounds good then. 
I'll merge it shortly.

PR #33, opened 2020-06-17
Dynamically load vulkan instead of statically link.

### LINKED ISSUE #17: Support loading vulkan functions from a dll
Currently Vk-bootrstrap assumes linking against the `vulkan-1.dl`l/`libvulkan.so`.
Make it so that by providing `vkGetInstanceProcAddr`, vk-bootstrap can call all the necessary functions.

charles-lunarg:
Make vk-bootstrap capable of loading the vulkan runtime and not need to
link against the library. This improves the usability of vk-bootstrap since
now you don't need the vulkan library on your system to build.

This commit also changes how SystemInfo works so as to allow the dynamic
vulkan loading.

Closes #17

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -57,10 +57,15 @@ auto inst_builder_ret = instance_builder
         .build();
 ```
 
-To query the available layers and extensions, use the `get_system_info()` function of `vkb::InstanceBuilder` to get a `SystemInfo` struct. It contains a `is_layer_available()` and `is_extension_available()` function to check for a layer or extensions before enabling it. It also has booleans for if the validation layers are present and if the VK_EXT_debug_utils extension is available.
+To query the available layers and extensions, get a `SystemInfo` struct from `SystemInfo::get_system_info()`. It contains a `is_layer_available()` and `is_extension_available()` function to check for a layer or extensions before enabling it. It also has booleans for if the validation layers are present and if the VK_EXT_debug_utils extension is available.
 
 ```cpp
-auto system_info = instance_builder.get_system_info();
+auto system_info_ret = vkb::SystemInfo.get_system_info();
+if (!system_info_ret) {
+    printf("%s\n", system_info_ret.error().message());
+    return -1;
+}
+auto system_info = system_info_ret.value();
 if (system_info.is_layer_available("VK_LAYER_LUNARG_api_dump")) {
     instance_builder.enable_layer("VK_LAYER_LUNARG_api_dump");
 }
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -16,18 +16,185 @@
 
 #include "VkBootstrap.h"
 
-#include <stdio.h>
-#include <string.h>
+#include <cstdio>
+#include <cstring>
+
+#if defined(_WIN32)
+#include <fcntl.h>
+#define NOMINMAX
+#include <windows.h>
+#endif // _WIN32
+
+#if defined(__linux__) || defined(__APPLE__)
+#include <dlfcn.h>
+#endif
+
+#include <mutex>
 
 namespace vkb {
 
 namespace detail {
 
-template <typename T>
-void get_inst_proc_addr (
-    T& out_ptr, const char* func_name, VkInstance instance, PFN_vkGetInstanceProcAddr ptr_vkGetInstanceProcAddr) {
-	out_ptr = reinterpret_cast<T> (ptr_vkGetInstanceProcAddr (instance, func_name));
-}
+class VulkanFunctions {
+	private:
+	std::mutex init_mutex;
+	struct VulkanLibrary {
+#if defined(__linux__) || defined(__APPLE__)
+		void* library;
+#elif defined(_WIN32)
+		HMODULE library;
+#endif
+		PFN_vkGetInstanceProcAddr ptr_vkGetInstanceProcAddr = VK_NULL_HANDLE;
+
+		VulkanLibrary () {
+#if defined(__linux__)
+			library = dlopen ("libvulkan.so.1", RTLD_NOW | RTLD_LOCAL);
+			if (!library) library = dlopen ("libvulkan.so", RTLD_NOW | RTLD_LOCAL);
+#elif defined(__APPLE__)
+			library = dlopen ("libvulkan.dylib", RTLD_NOW | RTLD_LOCAL);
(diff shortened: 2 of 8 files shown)


charles-lunarg:
There shouldn't be any need to change triangle.cpp's source code. The only functions being loaded are inside vk-bootstrap, it doesn't handle function loading for applications in general. 
That would be a lot of functionality to add, and also requires auto-gen, which I'm trying to avoid. Being able to pass in a `PFN_vkGetInstanceProcAddr` is what you'd use if you loaded vulkan by yourself or using another library.

src/VkBootstrap.h
```diff
@@ -199,8 +212,10 @@ void destroy_instance (Instance instance); // release instance resources
 
 class InstanceBuilder {
 	public:
-	// contains useful information about the available vulkan capabilities, like layers and instance extensions.
-	SystemInfo get_system_info () const;
+	// Default constructor, will load vulkan.
+	InstanceBuilder ();
+	// Optional: Can use your own PFN_vkGetInstanceProcAddr
+	InstanceBuilder (PFN_vkGetInstanceProcAddr fp_vkGetInstanceProcAddr);
```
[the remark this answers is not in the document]
charles-lunarg: Good catch, added explicit to the constructors.

src/VkBootstrap.cpp
```diff
@@ -16,18 +16,165 @@
...
+#if defined(__linux__)
+			library = dlopen ("libvulkan.so.1", RTLD_NOW | RTLD_LOCAL);
+			if (!library) library = dlopen ("libvulkan.so", RTLD_NOW | RTLD_LOCAL);
+#elif defined(__APPLE__)
+			library = dlopen ("libvulkan.dylib", RTLD_NOW | RTLD_LOCAL);
+			if (!library) library = dlopen ("libvulkan.1.dylib", RTLD_NOW | RTLD_LOCAL);
+#elif defined(_WIN32)
+			library = LoadLibrary (TEXT ("vulkan-1.dll"));
+#else
+			assert (false && "Unsupported platform");
```
martty, line 150: should be a #error

docs/getting_started.md
```diff
@@ -57,10 +57,13 @@ auto inst_builder_ret = instance_builder
         .build();
 ```
 
-To query the available layers and extensions, use the `get_system_info()` function of `vkb::InstanceBuilder` to get a `SystemInfo` struct. It contains a `is_layer_available()` and `is_extension_available()` function to check for a layer or extensions before enabling it. It also has booleans for if the validation layers are present and if the VK_EXT_debug_utils extension is available.
+To query the available layers and extensions, get a `SystemInfo` struct from `SystemInfo::get_system_info()`. It contains a `is_layer_available()` and `is_extension_available()` function to check for a layer or extensions before enabling it. It also has booleans for if the validation layers are present and if the VK_EXT_debug_utils extension is available.
 
 ```cpp
-auto system_info = instance_builder.get_system_info();
+auto system_info_ret = vkb::SystemInfo.get_system_info();
+if (!system_info_ret) return "Error: vulkan unavailable";
```
martty, line 64: Perhaps you wanted to printf this and return -1? It is just example code, but a bit weird.

src/VkBootstrap.cpp
```diff
@@ -16,18 +16,165 @@
...
+			FreeLibrary (library);
+#endif
+			library = 0;
+		}
+	};
+	VulkanLibrary& get_vulkan_library () {
+		static VulkanLibrary lib;
+		return lib;
+	}
+} // namespace detail
```
martty, line 176: This comment looks to be out of place.

src/VkBootstrap.cpp
```diff
@@ -16,18 +16,165 @@
...
+	}
+
+
+	bool load_vulkan (PFN_vkGetInstanceProcAddr fp_vkGetInstanceProcAddr = nullptr) {
+		if (fp_vkGetInstanceProcAddr != nullptr) {
+			ptr_vkGetInstanceProcAddr = fp_vkGetInstanceProcAddr;
+			return true;
+		} else {
+			auto lib = get_vulkan_library ();
+			ptr_vkGetInstanceProcAddr = lib.ptr_vkGetInstanceProcAddr;
```
martty, line 122: This function is almost thread-safe. If this function is thread-unsafe it makes everything invisibly thread-unsafe above.

src/VkBootstrap.cpp
```diff
@@ -16,18 +16,165 @@
...
+		get_inst_proc_addr (fp_vkDestroyImageView, "vkDestroyImageView");
+	}
+
+
+	bool load_vulkan (PFN_vkGetInstanceProcAddr fp_vkGetInstanceProcAddr = nullptr) {
+		if (fp_vkGetInstanceProcAddr != nullptr) {
+			ptr_vkGetInstanceProcAddr = fp_vkGetInstanceProcAddr;
+			return true;
+		} else {
+			auto lib = get_vulkan_library ();
```
martty, line 121: The copy here is not needed.

src/VkBootstrap.cpp
```diff
@@ -320,9 +467,27 @@ const char* to_string (SwapchainError err) {
 	}
 }
 
+bool init_vulkan_funcs (PFN_vkGetInstanceProcAddr fp_vkGetInstanceProcAddr) {
+	if (!detail::funcs.load_vulkan (fp_vkGetInstanceProcAddr)) return false;
+	detail::funcs.init_pre_instance_funcs ();
```
martty, line 472: This is not thread-safe. If it is intended to requery the function pointers, then maybe it would be better to store the pointers inside the SystemInfo. Otherwise using function statics or a mutex would make it threadsafe.

src/VkBootstrap.cpp
charles-lunarg: Good catch, fixed.

docs/getting_started.md
charles-lunarg: Fixed. Yes, its good to have the docs reflect actual usage of the library.

src/VkBootstrap.cpp
charles-lunarg: I didn't see a clear way to use function statics that would allow for data sharing between the initialization and getting of function pointers. Instead I opted to use mutexes, which should solve the problem readily, if at the cost of needing to use the `<mutex>` header. But well, even if users are very unlikely to do this, it's better to be safe than sorry.

src/VkBootstrap.cpp
charles-lunarg: Solved with a mutex.

PR #44, opened 2021-02-18
Address several issues in one fell swoop

### LINKED ISSUE #36: SwapchainBuilder doesn't get the allocation callbacks
Constructor forgets to pull the allocation callbacks out of the device.

### LINKED ISSUE #40: Specify default values for methods in `SwapchainBuilder` in their comments

### LINKED ISSUE #41: guard #define NOMINMAX
mingw64 defines this by itself in its os define, which vk-bootstrap pulls in via assert.h. An ifndef guard works.

### LINKED ISSUE #42: Expose `vkb::Result::m_error`
Please provide an accessor for `vkb::Result::m_error` so that I can pass the error directly to my error reporting

charles-lunarg:
Specifically these issues
Fixes #36 
Fixes #40 
Fixes #41 
Fixes #42

Also rewrote queue handling to use uint32_t everywhere with an added QUEUE_INDEX_MAX_VALUE as a sentinel. Removes excessive casting.

### DIFF
--- a/src/VkBootstrap.h
+++ b/src/VkBootstrap.h
@@ -165,17 +165,17 @@ const char* to_string (QueueError err);
 const char* to_string (DeviceError err);
 const char* to_string (SwapchainError err);
 
-// Gathers useful information about the available vulkan capabilities, like layers and instance extensions.
-// Use this for enabling features conditionally, ie if you would like an extension but can use a fallback if
-// it isn't supported but need to know if support is available first.
+// Gathers useful information about the available vulkan capabilities, like layers and instance
+// extensions. Use this for enabling features conditionally, ie if you would like an extension but
+// can use a fallback if it isn't supported but need to know if support is available first.
 struct SystemInfo {
 	private:
 	SystemInfo ();
 
 	public:
-    // Use get_system_info to create a SystemInfo struct. This is because loading vulkan could fail.
+	// Use get_system_info to create a SystemInfo struct. This is because loading vulkan could fail.
 	static detail::Result<SystemInfo> get_system_info ();
-	static detail::Result<SystemInfo>  get_system_info (PFN_vkGetInstanceProcAddr fp_vkGetInstanceProcAddr);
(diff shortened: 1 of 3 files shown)


src/VkBootstrap.h
```diff
@@ -93,11 +93,14 @@ template <typename T> class Result {
 	const T&& value () const&&   { assert (m_init); return std::move (m_value); }
 	T&&       value () &&        { assert (m_init); return std::move (m_value); }
 
+    // std::error_code associated with the error
     std::error_code error() const { assert (!m_init); return m_error.type; }
+    // optional VkResult that could of been produced due to the error
     VkResult vk_result() const { assert (!m_init); return m_error.vk_result; }
+    // Returns the internal struct that holds the errors
+    Error raw_error_struct() const { assert (!m_init); return m_error; }
```
DethRaid, line 101: I don't like that it's called `raw_error_struct`. That name implies it's an extra low level thing that shouldn't generally be used - to me, at least. I'd really like to use the name `error` for this method but that's already taken. What about `full_error`?

src/VkBootstrap.h
```diff
@@ -604,7 +620,7 @@ class SwapchainBuilder {
 
 	// Set the VkSwapchainCreateFlagBitsKHR.
 	SwapchainBuilder& set_create_flags (VkSwapchainCreateFlagBitsKHR create_flags);
-	// Set the transform to be applied, like a 90 degree rotation. Default is the current transform.
+	// Set the transform to be applied, like a 90 degree rotation. Default is the no transform.
```
DethRaid, line 623: "Default is no transform" not "default is **the** no transform"

DethRaid (changes requested):

src/VkBootstrap.h
charles-lunarg: I say raw because its an internal type to the library, not something that is meant to be directly used.
But with that said, full_error is more descriptive than 'raw_error_struct' anyhow. I'll fix it shortly

PR #47, opened 2021-03-01
test: Add windows builds to ci

charles-lunarg:


charles-lunarg:
Can you review the changes to CMakeLists.txt? 
Does it help your changes for conan?
@SpaceIm

SpaceIm:
I guess yes. We can now inject `Vulkan_INCLUDE_DIR` and avoid to patch CMakeLists.

When https://github.com/KhronosGroup/Vulkan-Headers/pull/112 will be merged, it will be possible to find specifically Vulkan-Headers with `find(VulkanHeaders)`, and link the target `Vulkan::Headers`.

charles-lunarg:
Would it be better to make the path an 'official' variable? So it'll use it if its provided, and if it isn't, try to find `Vulkan_INCLUDE_DIR`/manually download the headers. 
It helps for people integrating the library rather than having to read the CMakeLists.txt

SpaceIm:
Yes, I think that Vulkan-Loader and Vulkan-ValidationLayers have this kind of logic: https://github.com/KhronosGroup/Vulkan-Loader/blob/master/cmake/FindVulkanHeaders.cmake

charles-lunarg:
I added VK_BOOTSTRAP_VULKAN_HEADER_DIR as a build option. Sure, it is wordy, but it shouldn't have name collisions with anything else, which is more important. 

While the loader/validation layers FindHeaders.cmake has more functionality, I think its better to wait for Vulkan-Headers to get the proper cmake config and use that.

PR #53, opened 2021-03-30
Fixed a typo on README.md about dependencies

Lephar:
Changed "it's" to "its" on README.md, under the "Setting up vk-bootstrap" section. I believe it is meant that the library depends on the standard library of C++14, not that it is a standard library itself.


charles-lunarg (approved):
Yep, that definitely isn't right. And vk-bootstrap definitely is not a part of the standard library! Waiting on CI to finish before merging. Thanks!

PR #55, opened 2021-04-06
Check desired surface formats against feature flags

mehmetoguzderin:
This comparison is a simple addition that proves helpful when the developer intends to use swapchain obtained images for cases like storage purposes.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1603,14 +1603,20 @@ Result<SurfaceSupportDetails> query_surface_support_details(VkPhysicalDevice phy
 	return SurfaceSupportDetails{ capabilities, formats, present_modes };
 }
 
-VkSurfaceFormatKHR find_surface_format(std::vector<VkSurfaceFormatKHR> const& available_formats,
-    std::vector<VkSurfaceFormatKHR> const& desired_formats) {
+VkSurfaceFormatKHR find_surface_format(VkPhysicalDevice phys_device,
+    std::vector<VkSurfaceFormatKHR> const& available_formats,
+    std::vector<VkSurfaceFormatKHR> const& desired_formats,
+    VkImageUsageFlags usage_flags) {
 	for (auto const& desired_format : desired_formats) {
 		for (auto const& available_format : available_formats) {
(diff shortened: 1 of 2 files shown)


charles-lunarg:
This is a nice addition, and all the code looks idiomatic with the rest of the code base. 
At first I wondered why you only checked the `optimalTiling` but upon recollection and reading of the spec, I believe swapchain images must have a tiling mode of OPTIMAL_TILING (and they aren't buffers!) so there is no issue there.

Due to a cryptocurrency attack on Github Actions yesterday I disabled CI from running automatically, let me figure out how to get that happening and then we can see about merging this.

mehmetoguzderin:
@charles-lunarg thanks a lot for the paying attention, that attack sounds quite bad, maybe I should do the same to my repos too. Indeed, spec specifies that to be the case, I don't know how to link the exact table but it is in this section (for the future reader): https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#_wsi_swapchain

charles-lunarg:
I was only inconvenienced by the attack (20 spam email messages and realizing there are scummy people online). It certainly hurt github much more than I.
So I pushed a commit to master that allowed manually starting of jobs, but only on branches in the repo. I think if you rebase on master to get that change, you could run it in your fork and i'll just check that.
Yes, thanks for linking the spec.

mehmetoguzderin:
I think I am unable to (find the button to) run the job

charles-lunarg:
Yeah, either you have to setup actions on your fork (which I don't know how or what that actually means) or I need to figure out more of the github-actions to fix whatever is preventing this from occurring.

mehmetoguzderin:
Understood, PTAL here (in case it does the correct thing): https://github.com/mehmetoguzderin/vk-bootstrap/actions/runs/723878248

charles-lunarg:
Ah sweet. I think I can merge this, logic checks out and it passes the CI (in your fork).
I'm not sure what to do about running CI in the future, but for now me thinks asking contributors to manually run it should be fine.

Thanks for the contribution!

mehmetoguzderin:
Thanks for maintaining this great project!

PR #56, opened 2021-04-07
Fixes bug with 1.1/1.2 features not being appended to the pNext chain.

dangerfen:


charles-lunarg (approved):
Looks good to me!

Thanks for making my job easier and so I just have to hit the big green button to get this fixed.

PR #58, opened 2021-04-09
Added generic feature polling/matching/enabling.

dangerfen:
This PR is a bit of an overhaul of how features are handled internally and mainly affects DeviceBuilder and PhysicalDeviceSelector.

I added a crude reflection container that handles ownership/matching/destruction of said containers specific features struct instance. Vectors of these containers are used in both device creation and in physical device selection to verify that the specific feature fields requested in the feature structs are properly supported by a device.

Awhile ago I had to hack in some explicit checks for specific descriptor indexing features. I decided to try and make a generic impl that would support all range of feature structures so that we don't have to explicitly create a specific chunk of logic for every possible set of features there is.

Because it is generic, most of the pieces that were explicitly made around VkPhysicalDeviceVulkan11/12Features have been removed. set_required_features_11 and set_required_features_12 still exist, but they are now essentially just wrappers around add_required_features.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -844,7 +844,10 @@ std::vector<const char*> check_device_extension_support(
 }
 
 // clang-format off
-bool supports_features(VkPhysicalDeviceFeatures supported, VkPhysicalDeviceFeatures requested) {
+bool supports_features(VkPhysicalDeviceFeatures supported,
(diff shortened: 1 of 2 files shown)


dangerfen:
I need to take another look at `populate_device_details`  before this gets looked into. I realized that the field vectors are not being updated after the feature structs are updated via vkGetPhysicalDeviceFeatures2, which means that the matches will always pass.

PR #59, opened 2021-04-09
Added generic feature polling/enabling.

vibi-manx:
Recreated PR as breaking bug has been remedied.

This PR is a bit of an overhaul of how features are handled internally and mainly affects `DeviceBuilder` and `PhysicalDeviceSelector`.

I added a crude reflection container that handles ownership/matching/destruction of said containers specific features struct instance. Vectors of these containers are used in both device creation and in physical device selection to verify that the specific feature fields requested in the feature structs are properly supported by a device.

Awhile ago I had to hack in some explicit checks for specific descriptor indexing features. I decided to try and make a generic impl that would support all range of feature structures so that we don't have to explicitly create a specific chunk of logic for every possible set of features there is.

Because it is generic, most of the pieces that were explicitly made around `VkPhysicalDeviceVulkan11Features`/`VkPhysicalDeviceVulkan12Features` have been removed. `set_required_features_11` and `set_required_features_12` still exist, but they are now essentially just wrappers around `add_required_features`.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -908,7 +908,6 @@ bool supports_features(VkPhysicalDeviceFeatures supported,
     if (requested.inheritedQueries && !supported.inheritedQueries) return false;
 
     for(auto i = 0; i < extension_requested.size(); ++i) {
-        //auto res = extension_requested[i].match(extension_supported[i]);
         auto res = GenericFeaturesPNextNode::match(extension_requested[i], extension_supported[i]);
         if(!res) return false;
     }
@@ -1411,21 +1410,47 @@ detail::Result<Device> DeviceBuilder::build() const {
 		extensions.push_back({ VK_KHR_SWAPCHAIN_EXTENSION_NAME });
 
     bool has_phys_dev_features_2 = false;
+	bool user_defined_phys_dev_features_2 = false;
 	std::vector<VkBaseOutStructure*> final_pnext_chain;
+    VkDeviceCreateInfo device_create_info = {};
 
 #if defined(VK_API_VERSION_1_1)
+    for(auto& pnext : info.pNext_chain) {
+        if(pnext->sType == VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2) {
+            user_defined_phys_dev_features_2 = true;
+			break;
+		}
+    }
+
+    auto physical_device_extension_features_copy = physical_device.extended_features_chain;
     VkPhysicalDeviceFeatures2 local_features2{};
     local_features2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
-	auto physical_device_extension_features_copy = physical_device.extended_features_chain;
-    if (physical_device.instance_version >= VK_MAKE_VERSION(1, 1, 0)) {
-        has_phys_dev_features_2 = true;
-		for(auto& features_node : physical_device_extension_features_copy) {
-			final_pnext_chain.push_back(reinterpret_cast<VkBaseOutStructure*>(&features_node));
+
+	if(!user_defined_phys_dev_features_2) {
+		if (physical_device.instance_version >= VK_MAKE_VERSION(1, 1, 0)) {
+			local_features2.features = physical_device.features;
+            final_pnext_chain.push_back(reinterpret_cast<VkBaseOutStructure*>(&local_features2));
+			has_phys_dev_features_2 = true;
+			for (auto& features_node : physical_device_extension_features_copy) {
+				final_pnext_chain.push_back(reinterpret_cast<VkBaseOutStructure*>(&features_node));
+			}
 		}
+	}
+
+	if(!user_defined_phys_dev_features_2 && !has_phys_dev_features_2) {
+        device_create_info.pEnabledFeatures = &physical_device.features;
 	}
 #endif
 
-	VkDeviceCreateInfo device_create_info = {};
+    for(auto& pnext : info.pNext_chain) {
+        final_pnext_chain.push_back(pnext);
+    }
+
+    detail::setup_pNext_chain(device_create_info, final_pnext_chain);
+    for(auto& node : final_pnext_chain) {
+        assert(node->sType != VK_STRUCTURE_TYPE_APPLICATION_INFO);
+    }
+
 	device_create_info.sType = VK_STRUCTURE_TYPE_DEVICE_CREATE_INFO;
 	device_create_info.flags = info.flags;
 	device_create_info.queueCreateInfoCount = static_cast<uint32_t>(queueCreateInfos.size());
@@ -1433,29 +1458,9 @@ detail::Result<Device> DeviceBuilder::build() const {
 	device_create_info.enabledExtensionCount = static_cast<uint32_t>(extensions.size());
 	device_create_info.ppEnabledExtensionNames = extensions.data();
 
-	for(auto& pnext : info.pNext_chain) {
-        final_pnext_chain.push_back(pnext);
+	if(!final_pnext_chain.empty()) {
+        device_create_info.pNext = final_pnext_chain.front();
 	}
-
-	detail::setup_pNext_chain(device_create_info, final_pnext_chain);
-	for(auto& node : final_pnext_chain) {
-		assert(node->sType != VK_STRUCTURE_TYPE_APPLICATION_INFO);
-	}
-
-#if defined(VK_API_VERSION_1_1)
-	// VUID-VkDeviceCreateInfo-pNext-00373 - don't add pEnabledFeatures if the phys_dev_features_2 is present
-    if (has_phys_dev_features_2) {
-        device_create_info.pNext = &local_features2;
-		if(!final_pnext_chain.empty()) {
-            local_features2.pNext = final_pnext_chain.front();
-		}
-        device_create_info.pEnabledFeatures = nullptr;
-    } else {
-		device_create_info.pEnabledFeatures = &physical_device.features;
-	}
-#else
-	device_create_info.pEnabledFeatures = &physical_device.features;
-#endif
 
 	Device device;
 
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1015,9 +1015,12 @@ PhysicalDeviceSelector::PhysicalDeviceDesc PhysicalDeviceSelector::populate_devi
             prev = &extension;
         }
         if(desc.extension_features.size() > 0) {
-            desc.device_features2.pNext = &desc.extension_features[0].structure;
+            desc.device_features2.pNext = desc.extension_features[0].structure;
         }
         detail::vulkan_functions().fp_vkGetPhysicalDeviceFeatures2(phys_device, &desc.device_features2);
+        for(auto& extension : desc.extension_features) {
+			extension.update();
+		}
     }
 #endif
 	return desc;
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -31,6 +31,7 @@
 #include <dlfcn.h>
 #endif
 
+#include <cstdio>
 #include <mutex>
 
 namespace vkb {
@@ -1264,6 +1265,10 @@ PhysicalDeviceSelector& PhysicalDeviceSelector::set_desired_version(uint32_t maj
 	criteria.desired_version = VK_MAKE_VERSION(major, minor, 0);
 	return *this;
 }
+PhysicalDeviceSelector& PhysicalDeviceSelector::set_required_features(VkPhysicalDeviceFeatures const& features) {
+    criteria.required_features = features;
+    return *this;
+}
 #if defined(VK_API_VERSION_1_2)
 // Just calls add_required_features
 PhysicalDeviceSelector& PhysicalDeviceSelector::set_required_features_11(
@@ -1435,6 +1440,8 @@ detail::Result<Device> DeviceBuilder::build() const {
 				final_pnext_chain.push_back(reinterpret_cast<VkBaseOutStructure*>(&features_node));
 			}
 		}
+	} else {
+        printf("User provided VkPhysicalDeviceFeatures2 instance found in pNext chain. All requirements added via 'add_required_extension_features' will be ignored.");
 	}
 
 	if(!user_defined_phys_dev_features_2 && !has_phys_dev_features_2) {
@@ -1457,10 +1464,6 @@ detail::Result<Device> DeviceBuilder::build() const {
 	device_create_info.pQueueCreateInfos = queueCreateInfos.data();
 	device_create_info.enabledExtensionCount = static_cast<uint32_t>(extensions.size());
 	device_create_info.ppEnabledExtensionNames = extensions.data();
-
-	if(!final_pnext_chain.empty()) {
-        device_create_info.pNext = final_pnext_chain.front();
-	}
 
 	Device device;
 
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -31,7 +31,6 @@
 #include <dlfcn.h>
 #endif
 
-#include <cstdio>
 #include <mutex>
 
 namespace vkb {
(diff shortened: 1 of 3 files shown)


src/VkBootstrap.cpp
```diff
@@ -844,7 +844,10 @@ std::vector<const char*> check_device_extension_support(
 }
 
 // clang-format off
-bool supports_features(VkPhysicalDeviceFeatures supported, VkPhysicalDeviceFeatures requested) {
+bool supports_features(VkPhysicalDeviceFeatures supported,
+                       VkPhysicalDeviceFeatures requested,
+                       const std::vector<ExtensionFeatures>& extension_supported,
```
charles-lunarg, line 849: Gonna have to ask you to use east-const. (for consistency)
what that looks like.
There are multiple places this occurs fyi.
```cpp
std::vector<ExtensionFeatures> const& extension_supported,
```

src/VkBootstrap.h
```diff
@@ -419,16 +534,26 @@ class PhysicalDeviceSelector {
 	// Require a physical device that supports a (major, minor) version of vulkan.
 	PhysicalDeviceSelector& set_minimum_version(uint32_t major, uint32_t minor);
 
-	// Require a physical device which supports the features in VkPhysicalDeviceFeatures.
-	PhysicalDeviceSelector& set_required_features(VkPhysicalDeviceFeatures const& features);
```
charles-lunarg, line 423: Don't remove this function. (yes, there are other comments saying the same thing, one for each place it matters)

src/VkBootstrap.h
```diff
@@ -547,22 +670,13 @@ class DeviceBuilder {
...
-	}
-
 	// Provide custom allocation callbacks.
 	DeviceBuilder& set_allocation_callbacks(VkAllocationCallbacks* callbacks);
 
 	private:
 	PhysicalDevice physical_device;
 	struct DeviceInfo {
 		VkDeviceCreateFlags flags = 0;
-		std::vector<VkBaseOutStructure*> pNext_chain;
```
charles-lunarg, line 564: Same as before, don't remove this. It'll break existing code.

src/VkBootstrap.h
```diff
@@ -547,22 +670,13 @@ class DeviceBuilder {
 	// If a custom queue setup is provided, getting the queues and queue indexes is up to the application.
 	DeviceBuilder& custom_queue_setup(std::vector<CustomQueueDescription> queue_descriptions);
 
-	// Add a structure to the pNext chain of VkDeviceCreateInfo.
-	// The structure must be valid when DeviceBuilder::build() is called.
-	template <typename T> DeviceBuilder& add_pNext(T* structure) {
```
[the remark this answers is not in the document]
charles-lunarg: Not to mention that there *are* other use cases for the pNext chain than device features.

charles-lunarg (changes requested):
Okay first round of changes are a mix of 'what are you doing and why' alongside 'don't remove that, it breaks old code' with some small nitpicks mixed in.

I need to have the nitpicks and 'don't remove the functions' done so we can focus on the ExtensionFeatures implementation details. 

The other thing I need is to have some tests written which use this code. It already doesn't compile because of the breaking changes (which fixing made work again). With some tests, we can see the API usage as well as be something I can investigate the implementation in more detail.

src/VkBootstrap.h
```diff
@@ -195,6 +195,124 @@ struct SystemInfo {
 class InstanceBuilder;
 class PhysicalDeviceSelector;
 
+struct ExtensionFeatures {
+
+    using DeleteProc = void(*)(ExtensionFeatures&);
```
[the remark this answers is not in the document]
vibi-manx: The main reason is to try and keep the level of meta required for reflection as thin and as close to the type itself as possible.

If all three were to be static functions, they would need to be templated to ensure copying/updating/deleting is proper. This introduces a can of worms as the type/user/both would need a method of obtaining the type info in order to know which specialization to invoke. This is complicated even further for destruction.

Because all three are function pointers generated from lambdas built around T in the build function, all type info is baked into the procs beforehand, so no type info is needed when provoking copy/updating/destruction semantics. In the eyes of the type, handling these things is nothing more than calling a function pointer that does everything for it.

charles-lunarg:
I just had a thought. How are the sTypes's being initialized?
Are they left up to the user to insert correctly? 
as in:
```cpp
VkSomeFeaturesStruct s{};
s.sType = VK_STRUCTURE_TYPE_SOME_FEATURES_STRUCT;

//sType was filled in by the user.
physDevSelector.add_required_features(s);
```

charles-lunarg:
The issue is that I try to, when possible, prevent users from hurting themselves. 
Certainly, vulkan asks users to zero init structs and fill in sTypes in a bunch of places. However, what kind of error messages or crashes would they get if they didn't do this? Would validation layers catch it? Would you get memory errors due to the validation layers/driver reading the wrong data?
If I can at all help it, those are possibilities I want to make impossible when using the API.

vibi-manx:
In my experience validation output has been good at yelling at me when the sTypes are invalid. It even goes so far as to yell about incompatibilities (if you add both `VkPhysicalDevieVulkan12Features` and `VkPhysicalDeviceDescriptorIndexingFeaturesEXT` together for example).

I still believe that automating the sType set isn't something to worry to much about. Almost every aspect of the API expects properly formatted structs, and every tutorial resource goes over this in detail.

vibi-manx:
Requested changes have been made.

vibi-manx:
Tests were actually failing. Tests weren't showing as failed on VS for some reason but did on Clion. Fixed the break and now all tests pass.

charles-lunarg:
I have a potential solution for the 'need to fill out sTypes'.
Looking at the problem, there isn't a realistic method to get the sType to be *correct*. Adding another parameter for the sType doesn't guarantee that its correct and is annoying, creating a lookup table is going to bloat and require code gen and isn't forward compatible, which this extension currently supports. 

Instead, a suggestion I have gotten is to assert that the sType is not 0. This makes it so users will get yelled at if they don't set the sType. It doesn't make sure its the correct sType nor prevent uninitialized sType's. But it will catch stupid mistakes.

Putting this at the top of the function should suffice.
`assert(features.sType != 0 && "Features struct sType must be filled with the struct's corresponding VkStructureType enum");`

vibi-manx:
I had this same idea last night before sleeping and forgot about it. Good call out.

vibi-manx:
Just pushed a change that adds type `StructureContainer` to the `ExtensionFeatures` type which manages the features structure internally. It removes the need for any special update/copy/delete procs, greatly simplifying things.

src/VkBootstrap.h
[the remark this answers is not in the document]
vibi-manx: Closing as its no longer relevant since the removal of the function pointers and introduction of FeaturesContainer.

src/VkBootstrap.h
```diff
@@ -115,6 +115,34 @@ template <typename T> class Result {
...
+        memset(fields, 0, sizeof(fields));
+    }
+
+    VkStructureType sType = static_cast<VkStructureType>(0);
+    void* pNext = nullptr;
+    VkBool32 fields[256];
+
+    template <typename T>
+    void set(T const& features) {
+        GenericFeaturesPNextNode node;
```
charles-lunarg, line 131: Possible unused variable? I'm not getting a warning though

src/VkBootstrap.h
```diff
@@ -115,6 +115,34 @@ template <typename T> class Result {
...
+struct GenericFeaturesPNextNode {
+
+    GenericFeaturesPNextNode() : sType(static_cast<VkStructureType>(0)),
+                                 pNext(nullptr) {
+        memset(fields, 0, sizeof(fields));
+    }
+
+    VkStructureType sType = static_cast<VkStructureType>(0);
+    void* pNext = nullptr;
+    VkBool32 fields[256];
```
charles-lunarg, line 127: Would be good to create a static uint32_t member variable to store the number of fields.
```c
static const uint32_t USE_BETTER_NAME = 256;
VkBool32 fields[USE_BETTER_NAME];
```

src/VkBootstrap.h
```diff
@@ -359,6 +383,7 @@ struct PhysicalDevice {
 	uint32_t instance_version = VK_MAKE_VERSION(1, 0, 0);
 	std::vector<const char*> extensions_to_enable;
 	std::vector<VkQueueFamilyProperties> queue_families;
+    std::vector<detail::GenericFeaturesPNextNode> extended_features_chain;
```
charles-lunarg, line 386: Oooo this is a good name. "extended" fits very well.

src/VkBootstrap.cpp
```diff
@@ -900,75 +906,15 @@ bool supports_features(VkPhysicalDeviceFeatures supported, VkPhysicalDeviceFeatu
...
-    if(requested.bufferDeviceAddressMultiDevice && !supported.bufferDeviceAddressMultiDevice) return false;
-    if(requested.vulkanMemoryModel && !supported.vulkanMemoryModel) return false;
-    if(requested.vulkanMemoryModelDeviceScope && !supported.vulkanMemoryModelDeviceScope) return false;
-    if(requested.vulkanMemoryModelAvailabilityVisibilityChains && !supported.vulkanMemoryModelAvailabilityVisibilityChains) return false;
-    if(requested.shaderOutputViewportIndex && !supported.shaderOutputViewportIndex) return false;
-    if(requested.shaderOutputLayer && !supported.shaderOutputLayer) return false;
-    if(requested.subgroupBroadcastDynamicId && !supported.subgroupBroadcastDynamicId) return false;
+
+    for(auto i = 0; i < extension_requested.size(); ++i) {
+        //auto res = extension_requested[i].match(extension_supported[i]);
```
charles-lunarg, line 911: Commented out code should be removed. (I forget about this all the time, easy to miss.)

src/VkBootstrap.cpp
```diff
@@ -1467,70 +1410,61 @@ detail::Result<Device> DeviceBuilder::build() const {
...
+	detail::setup_pNext_chain(device_create_info, final_pnext_chain);
+	for(auto& node : final_pnext_chain) {
+		assert(node->sType != VK_STRUCTURE_TYPE_APPLICATION_INFO);
+	}
+
+#if defined(VK_API_VERSION_1_1)
 	// VUID-VkDeviceCreateInfo-pNext-00373 - don't add pEnabledFeatures if the phys_dev_features_2 is present
-	if (!has_phys_dev_features_2) {
+    if (has_phys_dev_features_2) {
+        device_create_info.pNext = &local_features2;
```
charles-lunarg, line 1448: Adding the features 2 struct to the pNext chain should be done with 
```c
final_pnext_chain.push_back(reinterpret_cast<VkBaseOutStructure*>(&local_features2));
```
That way there isn't any logic that needs to be correct but is hard to verify from a quick glance.

src/VkBootstrap.cpp
```diff
@@ -1467,70 +1410,61 @@ detail::Result<Device> DeviceBuilder::build() const {
...
 	device_create_info.queueCreateInfoCount = static_cast<uint32_t>(queueCreateInfos.size());
 	device_create_info.pQueueCreateInfos = queueCreateInfos.data();
 	device_create_info.enabledExtensionCount = static_cast<uint32_t>(extensions.size());
 	device_create_info.ppEnabledExtensionNames = extensions.data();
+
+	for(auto& pnext : info.pNext_chain) {
+        final_pnext_chain.push_back(pnext);
+	}
+
+	detail::setup_pNext_chain(device_create_info, final_pnext_chain);
```
charles-lunarg, line 1440: Do the pNext chain setup after the local_features2 struct has been added. Makes the logic easier.

charles-lunarg (commented):
Looking *really* good, only major thing is to look at the DeviceBuilder stuff, has the right bits but could be structured more simply by relying on the `final_pnext_chain` and the pNext chain setup function instead of manually doing it. (harder to mess up when there is less code.

src/VkBootstrap.h
vibi-manx: Yep. Unused and removed.

vibi-manx:
Changes have been made, and I also plugged in some logic for detecting if the user appended a `VkPhysicalDeviceFeatures2` instance to the pNext explicitly and and also added an assert for if they pass it to `add_required_extension_features`. Also noticed that the normal `VkPhysicalDeviceFeatures` was not being assigned to `VkPhysicalDeviceFeatures2` during chain setup. Thought about changing the name of `PhysicalDevice::features` to `PhysicalDevice::selected_features` as I noticed it doesn't reflect all of the features the device supports but is instead set to the value `criteria.required_features` during population, however that would result in an API break.

src/VkBootstrap.h
```diff
@@ -419,16 +445,31 @@ class PhysicalDeviceSelector {
...
+		       "Features struct sType must be filled with the struct's corresponding VkStructureType enum");
+		assert(features.sType != VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2 &&
+		       "Do not pass VkPhysicalDeviceFeatures2 as a required extension feature structure. An instance of this is managed internally for selection criteria and device creation.");
+		detail::GenericFeaturesPNextNode node;
+		node.set(features);
+        criteria.extended_features_chain.push_back(node);
+        return *this;
+    }
+#endif
+    PhysicalDeviceSelector& set_required_features(VkPhysicalDeviceFeatures const& features) {
```
charles-lunarg, line 462: Accidentally dropped the comment before this function.
`// Require a physical device which supports the features in VkPhysicalDeviceFeatures.`
Also would be nice to put the implementation back in the cpp file so the header is mostly only function declarations.

src/VkBootstrap.cpp
```diff
@@ -1467,70 +1409,67 @@ detail::Result<Device> DeviceBuilder::build() const {
...
 	device_create_info.queueCreateInfoCount = static_cast<uint32_t>(queueCreateInfos.size());
 	device_create_info.pQueueCreateInfos = queueCreateInfos.data();
 	device_create_info.enabledExtensionCount = static_cast<uint32_t>(extensions.size());
 	device_create_info.ppEnabledExtensionNames = extensions.data();
-	// VUID-VkDeviceCreateInfo-pNext-00373 - don't add pEnabledFeatures if the phys_dev_features_2 is present
-	if (!has_phys_dev_features_2) {
-		device_create_info.pEnabledFeatures = &physical_device.features;
-	}
 
+	if(!final_pnext_chain.empty()) {
```
charles-lunarg, line 1461: I am pretty sure this is redundant, `detail::setup_pNext_chain` connects all the structs together and puts the pointer to them inside the passed in struct (device_create_info)

src/VkBootstrap.h
```diff
@@ -419,16 +445,31 @@ class PhysicalDeviceSelector {
 	// Require a physical device that supports a (major, minor) version of vulkan.
 	PhysicalDeviceSelector& set_minimum_version(uint32_t major, uint32_t minor);
 
-	// Require a physical device which supports the features in VkPhysicalDeviceFeatures.
-	PhysicalDeviceSelector& set_required_features(VkPhysicalDeviceFeatures const& features);
-
+	// Require a physical device which supports a specific set of general/extension features.
+#if defined(VK_API_VERSION_1_1)
+    template <typename T>
+    PhysicalDeviceSelector& add_required_extension_features(T const& features) {
```
charles-lunarg, line 451: Just realized I don't think there is any logic to handle 'user added a features2 struct containing the VkPhysicalDeviceFeatures they want'. We use only the ones given by set_required_features. Afaik the code drops those features requested. I don't think its worth changing the code over, but possibly adding more info about it in the documentation would be good.
"// if the struct VkPhysicalDeviceFeatures2 is given to this function, the features member will be ignored. Use set_required_features instead."

charles-lunarg (commented):
One code change (for redundant code), and two documentation fixes. Close enough to taste...

src/VkBootstrap.h
charles-lunarg: And I only now realize that you  already have an assert here. That's even better. (Yes, I completely glossed over it in reviewing the code).

src/VkBootstrap.cpp
```diff
@@ -31,6 +31,7 @@
 #include <dlfcn.h>
 #endif
 
+#include <cstdio>
```
charles-lunarg, line 34: Not needed, it was included above on line 19.

charles-lunarg (commented):
One last thing....

charles-lunarg (approved):
After a week of back and forth, this finally is ready to be put to use. Now to wait for the people to find bugs we didn't even think to check for.

PR #61, opened 2021-04-16
CMake: add an option to enable warnings as errors

SpaceIm:
Warnings as errors are nice for developers of the project but not for consumers (simple user / package managers). Compilation could fail with new versions of compilers introducing new warnings.


charles-lunarg (approved):
Thanks, this is something I hadn't considered. 
Using warnings as errors forces me to fix my mistakes, but I can definitely see why that's not advantageous for users.

PR #69, opened 2021-05-06
Dispatch Table Generation

vibi-manx:
Included is a new `generate_dispatch.py` script which generates a `vkb::DispatchTable` type for device specific PFN's. The type is generated from `vk.xml` pulled directly from the Vulkan-Headers repo. PFN's are loaded by the type and are called via proxies that pass the tables `VkDevice` internally. `vkb::Device` has been added a function to return a table and some internals have been modified to store the `PFN_vkGetDeviceProcAddr`. A test has also been added.


src/VkDispatchTable.h
```diff
@@ -0,0 +1,4617 @@
...
+		fp_vkResetQueryPool = (PFN_vkResetQueryPool)procAddr(device, "vkResetQueryPool");
+		fp_vkGetSemaphoreCounterValue = (PFN_vkGetSemaphoreCounterValue)procAddr(device, "vkGetSemaphoreCounterValue");
+		fp_vkWaitSemaphores = (PFN_vkWaitSemaphores)procAddr(device, "vkWaitSemaphores");
+		fp_vkSignalSemaphore = (PFN_vkSignalSemaphore)procAddr(device, "vkSignalSemaphore");
+		fp_vkGetBufferDeviceAddress = (PFN_vkGetBufferDeviceAddress)procAddr(device, "vkGetBufferDeviceAddress");
+		fp_vkGetBufferOpaqueCaptureAddress = (PFN_vkGetBufferOpaqueCaptureAddress)procAddr(device, "vkGetBufferOpaqueCaptureAddress");
+		fp_vkGetDeviceMemoryOpaqueCaptureAddress = (PFN_vkGetDeviceMemoryOpaqueCaptureAddress)procAddr(device, "vkGetDeviceMemoryOpaqueCaptureAddress");
+#endif
+#if defined(VK_KHR_swapchain) && defined(VK_VERSION_1_0)
+#ifndef PFN_vkCreateSwapchainKHR_LOAD
```
charles-lunarg, line 186: What does the guards here accomplish? It seems that unless you define the guards, it will load the function. So by default it would load everything, which is fine. If the loader doesn't support a function, it'll return NULL so there is no worry about crashing when loading the function. The application should not call functions it didn't enable with the correct extensions, so its on them to do the right thing.

src/VkDispatchTable.h
```diff
@@ -0,0 +1,4617 @@
...
+
+#pragma once
+
+#include <vulkan/vulkan.h>
+
+namespace vkb {
+
+struct DispatchTable {
+	DispatchTable() = default;
+	DispatchTable(VkDevice const& device, PFN_vkGetDeviceProcAddr const& procAddr) : device(device) {
```
charles-lunarg, line 30: VkDevice is an opaque handle, eg a uint64_t. You don't need to pass by `const&`, pass by value is fine. Same goes for the proc addr, its is similarly small.

src/VkDispatchTable.h
```diff
@@ -0,0 +1,4617 @@
...
+		return fp_vkGetBufferDeviceAddress(device, pInfo);
+	}
+	uint64_t getBufferOpaqueCaptureAddress(const VkBufferDeviceAddressInfo* pInfo) {
+		return fp_vkGetBufferOpaqueCaptureAddress(device, pInfo);
+	}
+	uint64_t getDeviceMemoryOpaqueCaptureAddress(const VkDeviceMemoryOpaqueCaptureAddressInfo* pInfo) {
+		return fp_vkGetDeviceMemoryOpaqueCaptureAddress(device, pInfo);
+	}
+#if defined(VK_KHR_swapchain) && defined(VK_VERSION_1_0)
+#ifndef PFN_vkCreateSwapchainKHR_PROXY
```
charles-lunarg, line 1747: Why do these functions have guards for while the member functions above don't?
Similarly to when the functions are being loaded, these macro guards don't seem to be doing anything useful.

src/VkDispatchTable.h
```diff
@@ -0,0 +1,4617 @@
...
+	PFN_vkResetQueryPool fp_vkResetQueryPool = nullptr;
+	PFN_vkGetSemaphoreCounterValue fp_vkGetSemaphoreCounterValue = nullptr;
+	PFN_vkWaitSemaphores fp_vkWaitSemaphores = nullptr;
+	PFN_vkSignalSemaphore fp_vkSignalSemaphore = nullptr;
+	PFN_vkGetBufferDeviceAddress fp_vkGetBufferDeviceAddress = nullptr;
+	PFN_vkGetBufferOpaqueCaptureAddress fp_vkGetBufferOpaqueCaptureAddress = nullptr;
+	PFN_vkGetDeviceMemoryOpaqueCaptureAddress fp_vkGetDeviceMemoryOpaqueCaptureAddress = nullptr;
+#endif
+#if defined(VK_KHR_swapchain) && defined(VK_VERSION_1_0)
+#ifndef PFN_vkCreateSwapchainKHR_DECLARE
```
charles-lunarg, line 3499: Not sure what the macro guards here accomplish. Same question as the LOAD and PROXY guards elsewhere.

script/generate_dispatch.py
```diff
@@ -0,0 +1,387 @@
...
+header += 'struct DispatchTable {\n'
+header += '\tDispatchTable() = default;\n'
+header += '\tDispatchTable(VkDevice const& device, PFN_vkGetDeviceProcAddr const& procAddr) : device(device) {\n'
+
+proxy_definition = ''
+pfn_declaration = ''
+pfn_loading = ''
+
+for level in core_commands:
+	pfn_declaration += '#ifdef ' + level + '\n';
```
charles-lunarg, line 176: Consider using 'string interpolation' instead of adding everything together with +'s. This makes the code a lot more legible and fun to write.
https://www.programiz.com/python-programming/string-interpolation

charles-lunarg (changes requested):
Got some initial feedback for ya.

src/VkDispatchTable.h
```diff
@@ -0,0 +1,4617 @@
...
+#ifndef PFN_vkCmdEndVideoCodingKHR_LOAD
+#define PFN_vkCmdEndVideoCodingKHR_LOAD
+		fp_vkCmdEndVideoCodingKHR = (PFN_vkCmdEndVideoCodingKHR)procAddr(device, "vkCmdEndVideoCodingKHR");
+#endif
+#ifndef PFN_vkCmdControlVideoCodingKHR_LOAD
+#define PFN_vkCmdControlVideoCodingKHR_LOAD
+		fp_vkCmdControlVideoCodingKHR = (PFN_vkCmdControlVideoCodingKHR)procAddr(device, "vkCmdControlVideoCodingKHR");
+#endif
+#endif
+#if defined(VK_KHR_video_decode_queue) && defined(VK_VERSION_1_0)
```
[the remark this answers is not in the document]
vibi-manx: The version string is collected from the require nodes in extensions. I just set it to `VK_VERSION_1_0` when building the dictionary of extension commands so that there isn't any special branching needed in the final compilation.

src/VkDispatchTable.h
vibi-manx: The ones above are proxys build around 1.0 core, so they are guaranteed. Every other proxy is tied to an extension/require so it is guarded.

src/VkDispatchTable.h
vibi-manx: Some extension nodes had duplicate proto definitions from other extensions. `VK_KHR_swapchain` for example (line `11452`, cant link line directly because file is too large) has a require node with duplicate proto's from `VK_KHR_device_group`. The guards are to prevent multiple definitions from niche situations like this. I only see this happening twice throughout all of vk.xml, but you never know when things might change/be added.

src/VkDispatchTable.h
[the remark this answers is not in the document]
charles-lunarg: Then implement that logic in the autogen, it will make the resulting code much cleaner because then you don't have to rely on this 'guard in rare cases'. Autogen is powerful like that, you can work around potential problems in ways that with regular code is more obnoxious.

src/VkDispatchTable.h
vibi-manx: Extension PFN's were coming up as not-found unless the extension they were related to was also defined. One example is `VK_KHR_external_semaphore_win32` and the accompanying `PFN_vkImportSemaphoreWin32HandleKHR` which on my machine are showing as not defined.

charles-lunarg: The `#if defined(VK_KHR_swapchain) && defined(VK_VERSION_1_0)` line is fine, I'm specifically asking about `#ifndef PFN_vkCreateSwapchainKHR_LOAD` here. Needing the extension to be defined is weird however because `VK_KHR_swapchain` should always be 'defined' I though.

vibi-manx:
Going to do a bit of an overhaul based around the suggestions.

vibi-manx:
Generator script has been reworked. Unnecessary macros are no longer generated. Duplicate prototypes are now handled in autogen.

vibi-manx:
Some changes have been made to the way the dispatch table integrates into the core lib. The function added to the device type for creating a dispatch table has been removed, and instead now has a dispatch table instance as a member. The table is loaded by the builder with a new flagging function `populate_dispatch_table()`. The dispatch table type remains unchanged as far as initialization is concerned, so the "loading" is just an assignment with the needed constructor arguments to load. It remains decoupled from the core lib and can still be used standalone.

charles-lunarg (approved):
Looks great! Really appreciate all the work & back and forth that went on (some of which happened in channels other than github).

PR #77, opened 2021-06-07
Internal dispatch table usage for swapchain and device

vibi-manx:


charles-lunarg (approved):
Changes are good. This is what I should of done when I was adding function pointers, but lazyness won out and I opted for globals.

PR #78, opened 2021-06-08
Call KHR version of get phys dev properties if necessary

charles-lunarg:
Previously, vk-bootstrap didn't support calling vkGetPhyicalDeviceFeaturesKHR
on devices which support the extension. This makes vk-bootstrap more amenable
to 1.0 only hardware, like rasberry pi's.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -618,8 +618,9 @@ detail::Result<Instance> InstanceBuilder::build() const {
 	if (info.debug_callback != nullptr && system.debug_utils_available) {
 		extensions.push_back(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
 	}
-	if (detail::check_extension_supported(
-	        system.available_extensions, VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME)) {
+	bool supports_properties2_ext = detail::check_extension_supported(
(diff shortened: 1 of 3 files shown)


src/VkBootstrap.cpp
```diff
@@ -1724,27 +1749,25 @@ detail::Result<Swapchain> SwapchainBuilder::build() const {
...
 	Swapchain swapchain{};
-    PFN_vkCreateSwapchainKHR swapchain_create_proc;
-    detail::vulkan_functions().get_device_proc_addr(info.device, swapchain_create_proc, "vkCreateSwapchainKHR");
-    auto res = swapchain_create_proc(info.device, &swapchain_create_info, info.allocation_callbacks, &swapchain.swapchain);
-;	if (res != VK_SUCCESS) {
+	PFN_vkCreateSwapchainKHR swapchain_create_proc;
+	detail::vulkan_functions().get_device_proc_addr(info.device, swapchain_create_proc, "vkCreateSwapchainKHR");
+	auto res = swapchain_create_proc(
+	    info.device, &swapchain_create_info, info.allocation_callbacks, &swapchain.swapchain);
+	;
```
vibi-manx, line 1756: There is a stray semi here (that looks like it came from my previous commit)

vibi-manx (commented):
Aside from a stray semi everything looks good to me.

src/VkBootstrap.cpp
charles-lunarg: Thanks, I should do a separate PR that fixes some of the formatting issues I have with the code.

PR #79, opened 2021-06-13
Allow setting pUserData in Debug Messenger, move default callback to header file.

charles-lunarg:
The vulkan-loader implements the debug utils extension and always makes it available.
So this change makes it so that vk-bootstrap will always enable the extension and setup a
debug callback.

In addition, it is now possible to set pUserData if the user desires, which is nice for
custom debug callbacks

Lastly, the location of the default debug messenger function is now in the header file,
which makes it easier to find and possibly edit.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -16,7 +16,6 @@
 
 #include "VkBootstrap.h"
 
-#include <cstdio>
 #include <cstring>
 
 #if defined(_WIN32)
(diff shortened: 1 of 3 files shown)


vibi-manx (commented):
Not entirely comfortable making the debug messenger be enabled by default. While I understand the reasoning, it is normally recommended that the debug messenger only be used during development, and with the current changes there is no way to explicitly disable it when wanting to ship. If a function was added to allow disabling it, then we are just swapping a deprecated function for a new function that people have to update the codebases to flag when building in release.

That being said, I do like the addition of the user data pointer.

charles-lunarg:
> recommended that the debug messenger only be used during development
I would refute this as I've seen games turning on the info level for logging purposes when shipping to customers, that way they can analyze whats going on in the case of a crash.
However, I can see your reasoning in general, so I'll remove that addition. I originally just had the change be that the debug messenger is enabled when enabling validation layers, and I think that's valid behavior.

vibi-manx:
With that everything looks good to me.

PR #80, opened 2021-06-15
Small changes To GenericFeaturesPNextNode and DispatchTable generation

vibi-manx:
A small change has been made to GenericFeaturesPNextNode and the dispatch table now generates in UTF-8. This is mainly a requisite to a future PR I have planned.

### DIFF
--- a/script/generate_dispatch.py
+++ b/script/generate_dispatch.py
@@ -39,6 +39,7 @@ import sys
 import subprocess
 import pkg_resources
 import copy
+import codecs
 from string import Template
 
 installed = {pkg.key for pkg in pkg_resources.working_set}
@@ -287,7 +288,7 @@ body += '} // namespace vkb'
 
 header = license + info + body
 
-header_file = open("../src/VkBootstrapDispatch.h", "w")
+header_file = codecs.open("../src/VkBootstrapDispatch.h", "w", "utf-8")
 header_file.write(header)
 header_file.close();
 
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -40,7 +40,7 @@ GenericFeaturesPNextNode::GenericFeaturesPNextNode()
 : fields() {} // zero initializes the array of fields
 
 bool GenericFeaturesPNextNode::match(
-    GenericFeaturesPNextNode const& requested, GenericFeaturesPNextNode const& supported) {
+    GenericFeaturesPNextNode const& requested, GenericFeaturesPNextNode const& supported) noexcept {
 	assert(requested.sType == supported.sType && "Non-matching sTypes in features nodes!");
 	for (uint32_t i = 0; i < field_capacity; i++) {
 		if (requested.fields[i] && !supported.fields[i]) return false;
@@ -1138,6 +1138,7 @@ PhysicalDeviceSelector::PhysicalDeviceSelector(Instance const& instance) {
 
 detail::Result<PhysicalDevice> PhysicalDeviceSelector::select() const {
 
+#if !defined(NDEBUG)
 	// Validation
 	for(const auto& node : criteria.extended_features_chain) {
 		assert(node.sType != 0 && "Features struct sType must be filled with the struct's "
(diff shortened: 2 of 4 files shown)


src/VkBootstrap.cpp
```diff
@@ -1137,6 +1137,17 @@ PhysicalDeviceSelector::PhysicalDeviceSelector(Instance const& instance) {
 }
 
 detail::Result<PhysicalDevice> PhysicalDeviceSelector::select() const {
+
```
charles-lunarg, line 1140: I've found that in some compiler warnings report 'empty for loop' around these for loops. I at one point had a `#if !defined(NDEBUG)` surrounding them, but don't know where that code went.

src/VkBootstrap.h
```diff
@@ -121,7 +121,10 @@ struct GenericFeaturesPNextNode {
 
 	GenericFeaturesPNextNode();
```
charles-lunarg, line 122: Could be marked noexcept

src/VkBootstrap.h
```diff
@@ -121,7 +121,10 @@ struct GenericFeaturesPNextNode {
...
 
-	template <typename T> void set(T const& features) { *reinterpret_cast<T*>(this) = features; }
+	template <typename T>
+	GenericFeaturesPNextNode(T const& features) {
```
charles-lunarg, line 125: Could be marked noexcecpt

src/VkBootstrap.h
```diff
@@ -459,15 +462,7 @@ class PhysicalDeviceSelector {
...
-		assert(features.sType != 0 && "Features struct sType must be filled with the struct's "
-		                              "corresponding VkStructureType enum");
-		assert(
-		    features.sType != VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2 &&
-		    "Do not pass VkPhysicalDeviceFeatures2 as a required extension feature structure. An "
-		    "instance of this is managed internally for selection criteria and device creation.");
-		detail::GenericFeaturesPNextNode node;
-		node.set(features);
-		criteria.extended_features_chain.push_back(node);
+		criteria.extended_features_chain.push_back(features);
```
charles-lunarg, line 465: Not a criticism, just saying I like this code change :D

script/generate_dispatch.py
```diff
@@ -287,7 +288,7 @@
 
 header = license + info + body
 
-header_file = open("../src/VkBootstrapDispatch.h", "w")
+header_file = codecs.open("../src/VkBootstrapDispatch.h", "w", "utf-8")
```
charles-lunarg, line 291: Not thinking anything is wrong, but am curious to know what all this does and why it is necessary? (I can guess why but it'd be good to hear the reason for posterity's sake)

src/VkBootstrap.h
```diff
@@ -121,7 +121,10 @@ struct GenericFeaturesPNextNode {
 
 	GenericFeaturesPNextNode();
 
-	template <typename T> void set(T const& features) { *reinterpret_cast<T*>(this) = features; }
+	template <typename T>
+	GenericFeaturesPNextNode(T const& features) {
+		*reinterpret_cast<T*>(this) = features;
+	}
 
 	static bool match(GenericFeaturesPNextNode const& requested, GenericFeaturesPNextNode const& supported);
```
charles-lunarg, line 129: might as well ask for noexcept on eveything...
But the whole code file could probably get that treatment, so feel free to not add noexcept to the changed code, just say so (that way a single 'noexcept everything' can be made)

script/generate_dispatch.py
vibi-manx: The script was generating the source file in Windows-1252 encoding when I ran it, which makes me think that `open` defaults to the platforms default encoding when creating new files and not opening in binary. The change just makes sure the output file is in UTF-8.

vibi-manx:
NDEBUG guard added and noexcept flagged.

script/generate_dispatch.py
charles-lunarg: Ah that error, yeah this is a good idea then.

charles-lunarg (approved):
Thanks!

PR #87, opened 2021-09-06
fix minor typos and duplicate include paths

APokorny:
This is just a minor change to use vk-bootstrap with an install(export) setup in cmake.
The added "src" include path would contain CMAKE_SOURCE_DIR and that causes an error
message as it could not be exported into a cmake target config file.


charles-lunarg (approved):
Thanks!
`boostrap` is a typo I've done so many times it doesn't surprise me it snuck in here. And I believe the reason I added `src` to the include directory was to make code completion things happy. I'll merge it assuming it passes the CI run.

PR #88, opened 2021-09-10
Add user-defined conversion operators

spnda:
This adds [user-defined conversion](https://en.cppreference.com/w/cpp/language/cast_operator) operators to `Instance`, `Device`, `PhysicalDevice` and `Swapchain`. This basically means that one can simply pass any of the previously mentioned objects to any Vulkan function which accepts the Vulkan equivalent, say `VkInstance`, `VkDevice`, `VkPhysicalDevice`, and `VkSwapchainKHR` respectively. This would cut down on the need of getting the field each time and would be probably be a great enhancement, especially for Vulkan beginners.

I find it extremely painful to often be writing code like this (it also just seems a bit weird):
```cpp
vkb::Device device;
vkGetDeviceProcAddr(device.device, &addressInfo);
```
Now, to save just a few characters, I think it'd be nice if one could just write it like this:
```cpp
vkb::Device device;
vkGetDeviceProcAddr(device, &addressInfo);

### DIFF
--- a/example/triangle.cpp
+++ b/example/triangle.cpp
@@ -55,9 +55,9 @@ int device_initialization (Init& init) {
 	}
 	init.instance = instance_ret.value ();
 
-    init.vk_lib.init(init.instance.instance);
+    init.vk_lib.init(init.instance);
 
-	init.surface = create_surface_glfw (init.instance.instance, init.window);
+	init.surface = create_surface_glfw (init.instance, init.window);
 
 	vkb::PhysicalDeviceSelector phys_device_selector (init.instance);
 	auto phys_device_ret = phys_device_selector.set_surface (init.surface).select ();
@@ -74,7 +74,7 @@ int device_initialization (Init& init) {
 		return -1;
 	}
 	init.device = device_ret.value ();
-    init.vk_lib.init(init.device.device);
+    init.vk_lib.init(init.device);
 
 	return 0;
 }
@@ -146,7 +146,7 @@ int create_render_pass (Init& init, RenderData& data) {
 	render_pass_info.dependencyCount = 1;
 	render_pass_info.pDependencies = &dependency;
 
-	if (init->vkCreateRenderPass (init.device.device, &render_pass_info, nullptr, &data.render_pass) != VK_SUCCESS) {
+	if (init->vkCreateRenderPass (init.device, &render_pass_info, nullptr, &data.render_pass) != VK_SUCCESS) {
 		std::cout << "failed to create render pass\n";
 		return -1; // failed to create render pass!
 	}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by charles-lunarg** [comment]
This is a nice addition all in all.
I will ask if you have any comments on whether it should be explicit conversion (`explicit operator VkInstance();`) and what the impact on 32 bit, where the 'VkSwapchainKHR` handle will be a uint64_t value, not a pointer.

Additionally, I'd *love* to see some basic tests written with this, showcasing the usage and allowing the functionality to be tested.

**Comment by charles-lunarg** [comment]
And while I am suggesting modifications, putting a comment in the header file would be nice. Something like "This is a conversion operator, it allows using the vkb::Instance struct in places where VkInstance would have been used."

**Comment by spnda** [comment]
> whether it should be explicit conversion

I kinda don't think so, as it would mean one would have to write `VkDevice(device)` or `VkDevice vkDevice(device)`, instead of just `device` and `VkDevice vkDevice = device` respectively. That also has an interesting syntax but I prefer implicitly converting in this case. And I don't really see a real drawback of a user accidentally converting these types tbh.

> what the impact on 32 bit, where the 'VkSwapchainKHR` handle will be a uint64_t value, not a pointer.

I'm not exactly sure about this. Would this really matter though? I mean, from what I can see in vulkan_core.h, any 32-bit machine will have `VkSwapchainKHR` set to be `uint64_t`, see [here](https://github.com/KhronosGroup/Vulkan-Headers/blob/master/include/vulkan/vulkan_core.h#L57). Besides, any user would just be doing the same `swapchain.swapchain` in their code, which is effectively the same as this.

> Additionally, I'd love to see some basic tests written with this, showcasing the usage and allowing the functionality to be tested.

I guess I might just edit the current examples and use this syntax instead? I'll still add one separate test though.

> And while I am suggesting modifications, putting a comment in the header file would be nice.

As a doxygen comment on the functions? Or where do you mean exactly?

**Comment by charles-lunarg** [comment]
Sounds like I may have misremembered explicit conversions. Implicit should be fine.

The concern for 32 bit is actually related to the implicit conversion issues, though it should be fine now that I think about it since swapchain is the *only* type which will be 32 bit in vk-bootstrap. And if someone is compiling with 32 bit, they will have been bitten by this only when trying to use nondispatchable handles as type specifiers, eg overloading on the type. (which isn't likely with the types in use by vk-bootstrap).

The comment on comments was to put a short sentence in the VkBootstrap.h before the conversion operator declaration. Apologies for not being clear about that.
```c
struct Instance {
...
        // This is a conversion operator, it allows using the vkb::Instance struct in places where VkInstance would have been used.
 	operator VkInstance();
 	
```

spnda:
Nice. I've gone ahead and addressed everything and used the new syntax in the triangle example. I'm not sure how catch2 actually works so will have to look at that in a second. However, the triangle example works perfectly fine.

charles-lunarg (approved):
Changes look great!
Thanks for the submission :D

Issue #85, opened 2021-08-04
Replace with memcpy


charles-lunarg:
https://github.com/charles-lunarg/vk-bootstrap/blob/0408041547ad96ae05c6aae5c4af60f9b51bd6d9/src/VkBootstrap.h#L126

PR #89, opened 2021-09-11
Replacing reinterpret_cast with memcpy.

vibi-manx:
#85


charles-lunarg:
Awesome, should work just fine assuming the driver doesn't muck about with the padding bytes in the struct.

PR #94, opened 2021-10-21
Vulkan 1.1 compatibility

DethRaid:
This PR adds the KHR, EXT, or AMD suffix to VK1.2 types that were promoted from extensions. This lets vk-bookstrap support environments without Vulkan 1.2

This PR enables compatibility with Android NDK 22.0.7026061, the one I'm currently using. Other NDKs with different version of the Vulkan headers might not be compatible. However, as long as the environment has headers for at least Vulkan 1.1 it should be fine (assuming no extensions are at an unexpected revision)

Biggest potential change is that I had to separate our the KHR raytracing extension from the NV raytracing extension. This will hopefully not impact other users of vk-bs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,7 +80,7 @@ install(TARGETS vk-bootstrap
 
 option(VK_BOOTSTRAP_TEST "Test Vk-Bootstrap with glfw and Catch2" OFF)
 
-if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME OR VK_BOOTSTRAP_TEST)
+if (${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME} OR ${VK_BOOTSTRAP_TEST})
 
     add_subdirectory(ext)
     add_subdirectory(tests)
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -519,9 +519,13 @@ SystemInfo::SystemInfo() {
 		    detail::vulkan_functions().fp_vkEnumerateInstanceExtensionProperties,
 		    layer.layerName);
 		if (layer_extensions_ret == VK_SUCCESS) {
-			for (auto& ext : layer_extensions)
-				if (strcmp(ext.extensionName, VK_EXT_DEBUG_UTILS_EXTENSION_NAME) == 0)
+			for (auto& ext : layer_extensions) {
+				if (strcmp(ext.extensionName, VK_EXT_DEBUG_UTILS_EXTENSION_NAME) == 0) {
 					debug_utils_available = true;
+				}
+			}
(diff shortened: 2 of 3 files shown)


spnda:
I wonder about the actual reasoning and legitimacy behind this PR. As you're targeting Vulkan 1.1 (possibly Android?) wouldn't you also want to use Vulkan 1.1 headers (probably provided through Android SDK/NDK)? The changes you've made are mostly to function pointers which are guarded by a `#if (defined(VK_VERSION_1_2))` anyway.

If one is specifically targeting Vulkan 1.1, one could theoretically do `#undef VK_VERSION_1_2`. But for the generation of VulkanDispatchTable.h; I think it would most likely be best to have some preprocessor defines like `VKB_USE_VULKAN_1_1` to use the changes proposed by Deth and otherwise let it compile the current header. Using this, a Vulkan application using the 1.2 headers could be used to run on both 1.1 and 1.2, depending on the system.

`VK_AMD_draw_indirect_count` is not available on my NVIDIA card, however, `vkCmdDrawIndirectCount` is as it's in the core of Vulkan 1.2. So as this PR currently stands this might hurt Vulkan 1.2 users who depend on vkb and its dispatch table.

DethRaid:
> The changes you've made are mostly to function pointers which are guarded by a #if (defined(VK_VERSION_1_2)) anyway.

False. The guard are for if VK1.2 is defined _or if an extension is defined_. vk-bs uses the VK1.2 function no matter what - even if the extension function has a different name. This causes all kinds of compile errors with the Android VK1.1. headers. They define the extension functions but not the VK1.2 equivalents

charles-lunarg:
@spnda I have updated master to include what should be 1.1 compatibility. I would appreciate it if you could review and double check that it does indeed work.

spnda:
From a quick look it seems that with a1e58b4 it seems that my concerns have been fixed. It should also fix the 1.1 compatibility (possibly also 1.0 compatibility?) and actually fix the dispatch table as-is. Additionally, as `VK_VERSION_1_x` is deprecated, it might be better to use `VK_API_VERSION_1_x`?

charles-lunarg:
I noticed the usage of VK_API_VERSION/VK_MAKE_VERSION(). And yes it is deprecated, but it isn't easy to update as any 1.1/1.0 header wont have the newer versions. It would probably be better to copy/paste the definition into the vk-bootstrap code if we want to replace.

And thanks for looking over it. I didn't check if 1.0 works, I probably should. Its rather easy to do, uninstall any SDK on the system, edit the CMake FetchContent to pull 1.x.y, delete existing headers in the _deps folder, and thats all it takes. The XML currently generates of tip-of-main, so the xml generates with the most recent header version.

charles-lunarg:
I'm going to close the PR as I think the needs are adequately met.

PR #95, opened 2021-10-30
Append layer extensions to the list of available extensions

DethRaid:
Append the lists of extensions provided by Vulkan layers to the list of all extensions

This allows you to do things like use `VK_EXT_debug_utils` on Android. My phone's drivers only provide `VK_EXT_debug_report`, `VK_EXT_debug_utils` comes from the validation layer

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -519,9 +519,13 @@ SystemInfo::SystemInfo() {
 		    detail::vulkan_functions().fp_vkEnumerateInstanceExtensionProperties,
 		    layer.layerName);
 		if (layer_extensions_ret == VK_SUCCESS) {
-			for (auto& ext : layer_extensions)
-				if (strcmp(ext.extensionName, VK_EXT_DEBUG_UTILS_EXTENSION_NAME) == 0)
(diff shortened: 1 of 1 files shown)


charles-lunarg (commented):
What happens if a user enables an extension from a layer but doesn't enable the layer? vkb should try to catch that case. I can merge this PR and add that case later, unless you want to take a crack at it.

DethRaid:
I don't know how you or anyone else should handle that case. If the user is manually enabling an extension from a layer they aren't using, I'd expect VKB to be like "sorry, extension X not found"

charles-lunarg:
My issue is that combining the list of instance extensions with the list of extensions in layers is that the user will have no way to distinguish when they need to enable a layer vs not enabling the layer.

Alternatively, the layer could be enabled automatically if it implements the requested extension. But that's worse, as the same extension could be implemented by layers the application doesn't intend to enable, or by multiple layers (like for debug_utils).

Maybe there should be an explicit "enable extension from layer"? I don't like it, but it doesn't have the same drawbacks as the simpler solutions.

charles-lunarg:
I ended up taking this change (by creating #117) since it wouldn't require any additional effort on the original PR creator.

PR #96, opened 2021-10-30
Fix: Add VkBootstrapDispatch to includable files

spnda:
This change seems to be required for this project to work as a [vcpkg port](https://github.com/spnda/vcpkg/tree/master/ports/vk-bootstrap) as it otherwise can't find `VkBootstrapDispatch.h` in any include path. I'm no CMake expert so I'm not quite sure why because this project already does `target_include_directories`, but if it works it works. And I don't think it would break anything else.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ else ()
   endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
Yeah seems I forgot about updating the install stuff when adding this new file. Thank you for fixing it :)

PR #101, opened 2021-12-29
Fix crashing when calling GetPhysDevFeaturesKHR

### LINKED ISSUE #99: [Linux] segfault on vkGetPhysicalDeviceFeatures2KHR
As seen in the below snippet, vk-bootstrap checks if the Vulkan API version is newer than 1.1 and if so, uses the core version of `vkGetPhysicalDeviceFeatures2` so it doesn't have to enable the extension. However, on my Linux machine, it somehow doesn't go that route and and tries calling `vkGetPhysicalDeviceFeatures2KHR`, which is `nullptr` and therefore causes a segfault. Commenting out that line and the application runs fine. My instance info API version is `1.2.0` and my device reports as `1.2.186`. This doesn't happen on my Windows machine, which leads me to believe it's some Linux driver specific issue possibly?

https://github.com/charles-lunarg/vk-bootstrap/blob/142986cdb767037118b687387b097ff6b3e7fe7d/src/VkBootstrap.cpp#L1047-L1051

charles-lunarg:
vk-bootstrap would attempt to call vkGetPhysDevFeaturesKHR without enabling the appropriate extension first.

Fixes #99

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -620,7 +620,7 @@ detail::Result<Instance> InstanceBuilder::build() const {
 	bool supports_properties2_ext = detail::check_extension_supported(
 	    system.available_extensions, VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
 
-	if (supports_properties2_ext && api_version < VK_API_VERSION_1_1) {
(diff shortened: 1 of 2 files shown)


jackie-fuchs:
Everything looks good to me.

Issue #102, opened 2022-01-03
Allow selecting a device based on the name of a Physical Device


charles-lunarg:
Currently, there is no easy way to get the list of names of the available physical devices nor select a physical device based on that list.

Suggested improvements:
1. Create a function to query the names of the physical devices (allowing applications to present that list to the user
2. Add a 'select' like function to `vkb::PhysicalDeviceSelector` that returns a list of all available physical devices which meet the criteria. This lets apps get all the physical devices which meet the requirements.
3. Add a name as a criteria for selecting a physical device.

playmer:
Regarding 2: Should partially suitable devices be returned? Or should they only be returned if no fully suitable devices be returned? I think I lean towards the second, since that would allow the user/engine more flexibility. Maybe it could be a bool parameter for the function?

charles-lunarg:
Partially suitable devices are a poor solution for the 'I must have X, I would like Y' because there isn't a clear way to find out if Y is available.
I would like to not support that here, or at least make it more explicit like with a bool parameter. `get_suitable_devices(bool return_partially_suitable)` and the like.

charles-lunarg:
FYI, a partially suitable physical devices are from the desire to support apps where a known base spec is required but there is support for optional features, like raytracing. However, the API doesn't make it at all clear which features a given physical device supports. Like, you can be sure the device supports the required features but absolutely no clue about the optional stuff. 
I'm not asking for a solution to this issue, but that I'm open to discussion & ideas on how to improve the situation.

playmer:
Yeah I didn't even know partially suitable devices were a thing in the library until I started poking at the select function :P so I don't yet have an opinion or idea, but I'll try to think about it.

playmer:
So I've been factoring stuff out of `get_suitable_devices(bool)` and `select()` and rather than having a function to return just the device names, maybe we should just make public the `PhysicalDeviceDesc` struct and a function to get all of those since I already have that as a private function since both need it?

jackie-fuchs:
I am confused as to why all of this needs to be refactored out when the name could just be added as a string to the PhysicalDevice struct that is already returned.

playmer:
Well I refactored it out more to reduce code duplication between the two functions. But that would work as well of course.

playmer:
The other reason was to not bring in `std::string` as a dependency on the header.

charles-lunarg:
[...]

PR #104, opened 2022-01-04
Physical device selection by name.

playmer:
So this is an attempt at tackling #102. 

- [x] Create a function to query the names of the physical devices (allowing applications to present that list to the user
- [x] Add a 'select' like function (named `PhysicalDeviceSelector::get_suitable_devices()`) to vkb::PhysicalDeviceSelector that returns a list of all available physical devices which meet the criteria. This lets apps get all the physical devices which meet the requirements.
- [x] Add a name as a criteria (named`PhysicalDeviceSelector::set_required_device_name()`) for selecting a physical device.

Since `PhysicalDeviceSelector::get_suitable_devices()` is similar to `PhysicalDeviceSelector::select()` I factored out some of each of them into functions to do some code dedupe. Happy to not do that or do some more (the validation stuff at the beginning of the functions could maybe be done...) depending on taste. 

Let me know if there's anything that needs adjusting!

One thing I noticed when testing was that I did actually need to pass the suitability flag as true when listing all GPUs because an iGPU is determined to only have partial support by default and there's no way to make any GPU return `Suitable::yes` (when that's the only issue). `SelectionCriteria::allow_any_type` feels like this should return a yes here, but we might need a different bool instead to see if we check this at all.

Here's the dance I added in my little app framework to test out using these features on my laptop with both a dGPU and iGPU. 

```cpp
///////////////////////////////////////
// Select Physical Device
vkb::PhysicalDeviceSelector phys_device_selector(mInstance);
phys_device_selector.set_surface(mSurface);

// All the devices we could support. Right now we just print them out, but we may wish
// to eventually pipe this up to the application layer.
{
  auto potential_physical_devices_return = phys_device_selector.get_suitable_devices(true);
  if (!potential_physical_devices_return) 
  {
    printf("Failed to select Vulkan Physical Device. Error: %s\n", potential_physical_devices_return.error().message().c_str());
  }

  for (auto& physical_device : potential_physical_devices_return.value())
  {
    printf("Physical Device Name: %s\n", physical_device.properties.deviceName);
  }
}

// We let the selector pick us the preferred GPU without user input, in case
// there's some issue with the user provided GPU.
{
  auto physical_device_selector_return = phys_device_selector.select();

  if (!physical_device_selector_return) 
  {
    // We return out because there's really nothing we can do at this point, there's not a 
    // single suitable GPU on this system.
    printf("Failed to select Vulkan Physical Device. Error: %s\n", physical_device_selector_return.error().message().c_str());
    return;
  }
  else
  {
    mPhysicalDevice = physical_device_selector_return.value();
  }
}

// Finally, we attempt to let the user select a particular GPU
if (nullptr != aPreferredGpu)
{
  phys_device_selector.set_required_device_name((char const*)aPreferredGpu);

  auto physical_device_selector_return = phys_device_selector.select();
  if (!physical_device_selector_return) 
  {
    printf("Failed to select Vulkan Physical Device named \"%s\", falling back to \"%s\". Error: %s\n", 
      aPreferredGpu,
      mPhysicalDevice.properties.deviceName,
      physical_device_selector_return.error().message().c_str());
  }
  else
  {
    mPhysicalDevice = physical_device_selector_return.value();
  }
}
```

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1069,6 +1069,8 @@ PhysicalDeviceSelector::Suitable PhysicalDeviceSelector::is_device_suitable(Phys
 	if (criteria.required_version > pd.device_properties.apiVersion) return Suitable::no;
 	if (criteria.desired_version > pd.device_properties.apiVersion) suitable = Suitable::partial;
 
+	bool name = strcmp(criteria.desired_device_name, pd.device_properties.deviceName) == 0;
+
 	bool dedicated_compute = detail::get_dedicated_queue_index(pd.queue_families,
 	                             VK_QUEUE_COMPUTE_BIT,
 	                             VK_QUEUE_TRANSFER_BIT) != detail::QUEUE_INDEX_MAX_VALUE;
@@ -1086,6 +1088,7 @@ PhysicalDeviceSelector::Suitable PhysicalDeviceSelector::is_device_suitable(Phys
 	    detail::get_present_queue_index(pd.phys_device, instance_info.surface, pd.queue_families) !=
 	    detail::QUEUE_INDEX_MAX_VALUE;
 
+	if (criteria.require_device_name && !name) return Suitable::no;
 	if (criteria.require_dedicated_compute_queue && !dedicated_compute) return Suitable::no;
 	if (criteria.require_dedicated_transfer_queue && !dedicated_transfer) return Suitable::no;
 	if (criteria.require_separate_compute_queue && !separate_compute) return Suitable::no;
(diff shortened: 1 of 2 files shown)


jackie-fuchs:
`set_required_device_name` seems redundant when the `vkb::PhysicalDevice` already contains the name and is returned, plus it introduces the possibility of people creating issues related to not getting back any devices because they either misspelled the required name or did not spell it out the exact way Vulkan reports it. The returned vector of `vkb::PhysicalDevice` should be enough on its own because the user can easily peak it and see if there are any misspellings/slight differences that need to be corrected instead of having them dive into the source in order to peak the passed value and the one assigned in the `VkPhysicalDeviceProperties`.

src/VkBootstrap.h
```diff
@@ -489,6 +492,9 @@ class PhysicalDeviceSelector {
 	// Prefer a memory heap from VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT with `size` memory available.
 	PhysicalDeviceSelector& desired_device_memory_size(VkDeviceSize size);
 
+	// Require a physical device which has a specific name.
+	PhysicalDeviceSelector& set_required_device_name(const char* extension);
```
charles-lunarg, line 496: Parameter name is `extension` which is weird.

src/VkBootstrap.h
```diff
@@ -440,6 +440,7 @@ struct PhysicalDevice {
 	operator VkPhysicalDevice() const;
 
 	private:
+	char device_name[VK_MAX_PHYSICAL_DEVICE_NAME_SIZE];
```
charles-lunarg, line 443: Possibly left in by accident? Since the name is in `VkPhysicalDeviceProperties` and this is a private variable.

playmer:
Hmm, I don't really agree, while it is more finnicky than other criteria, I don't think it's particularly different in that regard. I would expect that a Developer would provide a list via UI, populated using `PhysicalDeviceSelector::get_suitable_devices()` and then populate the string in a configuration file themselves such that the User is not themselves writing it out. (Although for now I am just copy/pasting the value for testing.) 

I admit that beginners might run into your scenario (and thus I should probably write an extended comment describing the expectation of its use), but I don't think that's a reason to get rid of the feature.

src/VkBootstrap.h
```diff
@@ -557,23 +563,29 @@ class PhysicalDeviceSelector {
...
+		bool require_device_name = false;
 		bool require_present = true;
 		bool require_dedicated_transfer_queue = false;
 		bool require_dedicated_compute_queue = false;
 		bool require_separate_transfer_queue = false;
 		bool require_separate_compute_queue = false;
 		VkDeviceSize required_mem_size = 0;
 		VkDeviceSize desired_mem_size = 0;
-
+		const char* desired_device_name = "";
```
charles-lunarg, line 588: Probably should read `requested_device_name` since its a 'required' value not a desired.

src/VkBootstrap.cpp
```diff
@@ -1283,6 +1365,11 @@ PhysicalDeviceSelector& PhysicalDeviceSelector::desired_device_memory_size(VkDev
 	criteria.desired_mem_size = size;
 	return *this;
 }
+PhysicalDeviceSelector& PhysicalDeviceSelector::set_required_device_name(const char* extension) {
```
charles-lunarg, line 1368: Parameter name is wrong here too.

jackie-fuchs:
I can see where you are coming from, but the exact same effect can be achieved with a simple comparison against the returned `vkb::PhysicalDevice`'s from `get_suitable_devices` and the value that was read from said configuration file. This doesn't carry the weight of potential fallacy and keeps the internals leaner.

charles-lunarg (commented):
Not finished with the review, but did find some nit picks.

playmer:
Agreed that it's not a huge blocker either way, I'll leave it to Charles make a decision either way. I think it's nice to have in there, but certainly not needed.

charles-lunarg:
Thanks for the work on this branch! Gonna close it now that the other branch has all the same functionality.

PR #106, opened 2022-01-05
CMake: export vk-bootstrap and dependent targets

### LINKED ISSUE #105: Export `vk-bootstrap` targets in CMake for easy upstream consumption
While `vk-bootstrap` installs its public targets, it does not yet export them into an export-set, which would be required by upstream projects trying to export their dependencies. Since `vk-bootstrap` is aware of its dependencies, it should export them like so:
```cmake
install(TARGETS vk-bootstrap vk-bootstrap-compiler-warnings vk-bootstrap-vulkan-headers
        EXPORT vk-bootstrap-targets
        # ...
)
```

Then upstream projects will be able to simply `install(EXPORT vk-bootstrap ...)` as they desire.

Note: ideally `vk-bootstrap` should also install a package config file (to support `find_package`), but this would at least enable adding these targets to an existing package.

karnkaul:
Enable upstream scripts to install the `vk-boostrap-targets` export-set directly.
Close #105

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,8 @@ target_compile_features(vk-bootstrap PUBLIC cxx_std_14)
 
 include(GNUInstallDirs)
 install(FILES src/VkBootstrap.h src/VkBootstrapDispatch.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
-install(TARGETS vk-bootstrap
+install(TARGETS vk-bootstrap vk-bootstrap-compiler-warnings vk-bootstrap-vulkan-headers
+        EXPORT vk-bootstrap-targets
         RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
         LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
         ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})


charles-lunarg:
Since the two targets added are a 'header only' and a 'compiler options only', how does this affect what the outputs of installing the library are? 
Or are these changes so that downstream packagers don't have to do as much work. 
I am unfamiliar with the rules around packaging & using cmake in this context.
Also greatly appreciate raising an issue then creating a PR to solve it :)

karnkaul:
There shouldn't be any changes to the files that will be installed; this is the result of installing with these changes:
```
$ cmake --install out --prefix=out/install
-- Install configuration: "Debug"
-- Installing: /.../vk-bootstrap/out/install/include/VkBootstrap.h
-- Installing: /.../vk-bootstrap/out/install/include/VkBootstrapDispatch.h
-- Installing: /.../vk-bootstrap/out/install/lib/libvk-bootstrap.a

$ ls -R out/install
out/install:
include  lib

out/install/include:
VkBootstrapDispatch.h  VkBootstrap.h

out/install/lib:
libvk-bootstrap.a
```

It's essentially just creating an export-set within CMake that is aware of all the targets to then import `vk-bootstrap` as an `IMPORTED` library that `target_link_libraries` etc can be used on. If the set is not used by any other CMake files in the tree, I don't think there will be any noticeable difference in the outputs.

Also, my pleasure! Glad to be able to contribute! :D

charles-lunarg:
Ah then I don't see a reason to reject the PR.

PR #116, opened 2022-02-09
Use own Vulkan version macros

spnda:
The macros were used inconsistently used; sometimes `VK_API_VERSION_x_y`, sometimes `VK_VERSION_x_y`. I did not change the dispatch table, but they definitions could be re-used there too, as previously suggested in https://github.com/charles-lunarg/vk-bootstrap/issues/94#issuecomment-1031361373. And it should fix using older header versions where the `_API_` macros don't exist yet.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -570,14 +570,14 @@ detail::Result<Instance> InstanceBuilder::build() const {
 	if (!sys_info_ret) return sys_info_ret.error();
 	auto system = sys_info_ret.value();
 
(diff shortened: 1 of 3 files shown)


charles-lunarg:
Eh, its a good change in the end. When I saw the PR title I thought "I bet I'm going to have to ask for X, Y, and Z changes", except, it already has every one of those things. No harm taking it, and it is more inline with where the Vulkan API headers are moving.

charles-lunarg (approved):

PR #120, opened 2022-02-19
Fix version macro not selecting VK_MAKE_VERSION

forenoonwatch:
The version macro selector was looking for the constant VK_API_VERSION instead of the correct macro VK_MAKE_VERSION. Replaced the preprocessors accordingly.


spnda:
oops...

PR #121, opened 2022-02-21
Allow desiring higher api versions

Danderaion:
Allows applications to desire_api_version that is
higher than available instance version.

It is possible to use PhysicalDevices with newer api version as long as your Instance is at least 1.1.
However application **must** not use newer api features if ApplicationInfo.apiVersion is set to lower value. ( [Vk spec Chap4. #VkApplicationInfo](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/chap4.html#VkApplicationInfo) )

This PR allows one to create instance with lower version while keeping ApplicationInfo::apiVersion at desired value

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -576,10 +576,9 @@ detail::Result<Instance> InstanceBuilder::build() const {
 	if (!sys_info_ret) return sys_info_ret.error();
 	auto system = sys_info_ret.value();
 
-	uint32_t api_version = VKB_VK_API_VERSION_1_0;
 	uint32_t instance_version = VKB_VK_API_VERSION_1_0;
 
-	if (info.required_api_version > VKB_VK_API_VERSION_1_0 ||
+	if (info.minimum_instance_version > VKB_VK_API_VERSION_1_0 || info.required_api_version > VKB_VK_API_VERSION_1_0 ||
 	    info.desired_api_version > VKB_VK_API_VERSION_1_0) {
 		PFN_vkEnumerateInstanceVersion pfn_vkEnumerateInstanceVersion =
 		    detail::vulkan_functions().fp_vkEnumerateInstanceVersion;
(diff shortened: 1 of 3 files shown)


charles-lunarg:
Looking at the code, it might make sense to simply add an explicit instance version field. That way its always clear what the application is setting. 
This would entail
* Adding instance api fields/functions
* Renaming existing version fields/functions to device api version (or add new device api fields/functions but deprecate the existing ones)
* Forwarding that info to the physical device selector

This PR definitely allows for what you need, but it muddies the water of what 'desired vs required' means, and thats something I've been meaning to remove.

Danderaion:
Hmm, so what if something like:

- Deprecate current `desire_api_version` (or keep it around working like this patch. i.e. set `apiVersion` to that, but don't forward)
- add `required/minimum_instance_version` field to `vkb::InstanceBuilder`
- add `target/required_api_version` field to `vkb::Instance`
- `require_api_version` stays the same. This is also forwarded to `PhysicalDeviceSelector` like currently. (in a new field)
- Add `set_minimum_instance_version` which **explicitly** allows lower instance version when targeting higher version (like `set_minimum_version` on `PhysicalDeviceSelector`)
- `vkb::Instance` still passes on actual instance version onwards to `PhysicalDeviceSelector` which doesn't need any new fields. Instance version goes to `instance_info.version` and `target_api_version` goes to `criteria.required/desired_version`

This would require users to opt-in to lower instance versions so they don't get suprised when instance version is lower than expected while still allowing usage of vkb on systems with 1.2 (or even 1.1) loader and 1.3 driver

Also... should `PhysicalDeviceSelector::set_desired_version` also be deprecated? Since you shouldn't ask for anything beyond your `VkApplicationInfo::apiVersion` which is set during `InstanceBuilder::build` and then forwarded to `PhysicalDeviceSelector`.

charles-lunarg:
Apologies for not getting back to you sooner.
I think I have what I want to happen.

* `InstanceBuilder::require_api_version`: (applies to both instance & device). This is the 'stoopid simple' option. 

* `InstanceBuilder::desire_api_version` -> `set_minimum_instance_version`. Only fails if the available instance version is less. Deprecate `desire_api_version` (keep it but comment that its going away)

* `PhysicalDeviceSelector::set_minimum_version` stays put.

* `PhysicalDeviceSelector::set_desired_version` is deprecated. (keep it but comment that its going away)

This is probably exactly what you had intended but with wording changes. I needed to write it out so I understood what was happening. 

Also: You can change this PR or ask me to do it, you don't need to provide free labor to this. :)

Danderaion:
Those names look good to me.
I can update the PR probably later today

Danderaion:
I might have forgotten to actually do it on Friday, but better late than never, I guess.

charles-lunarg (approved):
Perfect!
Thanks for putting together these changes. Definitely is more useful and straight forward to use now

PR #130, opened 2022-04-21
Fix best practices warning when in VK 1.1 or higher

nickclark2016:
When using an API version other than VK 1.0 (read 1.1 and higher), we should use the functionality in core rather than the extension for getting the device physical properties.  This resolves the output produced by the best practices layer.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -610,7 +610,7 @@ detail::Result<Instance> InstanceBuilder::build() const {
 	bool supports_properties2_ext =
 	    detail::check_extension_supported(system.available_extensions, VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
 
-	if (supports_properties2_ext) {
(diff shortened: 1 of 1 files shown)


src/VkBootstrap.cpp
```diff
@@ -610,7 +610,7 @@ detail::Result<Instance> InstanceBuilder::build() const {
 	bool supports_properties2_ext =
 	    detail::check_extension_supported(system.available_extensions, VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
 
-	if (supports_properties2_ext) {
+	if (supports_properties2_ext && api_version == VK_API_VERSION_1_0) {
```
charles-lunarg, line 613: Should be `api_version < VKB_VK_API_VERSION_1_1`
This way anyone setting a patch version isn't screwed and it uses the macro defines which correctly use the right version macro.

charles-lunarg (changes requested):
Can't merge it straight out, but very much a good fix.

charles-lunarg (approved):
Woot woot!

PR #135, opened 2022-04-30
Fix c543dc1: portability_enumeration flag was not available before 1.3.208

spnda:
c543dc1 introduced support for `VK_KHR_portability_enumeration`, but macro guard around the usage of `VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR` seems to have been forgotten, which made the program not compile for older header versions.


charles-lunarg (approved):
Thank you! Definitely slipped my notice when creating the change.

Harder to test these kinds of code changes when CI always uses the latest headers.

Issue #137, opened 2022-05-15
Swapchain builder: image_count is set to minImageCount+1; consider providing a way to override it?


yoanlcq:
Hi,

In my engine I'd like the ability to choose how many images a swapchain should have, since I assume it involves a trade-off between memory consumption and the probability that the next image is ready for use as we acquire it.

The relevant part of the code is :

```cpp
detail::Result<Swapchain> SwapchainBuilder::build() const {
        // ...
	uint32_t image_count = surface_support.capabilities.minImageCount + 1;
	if (surface_support.capabilities.maxImageCount > 0 && image_count > surface_support.capabilities.maxImageCount) {
		image_count = surface_support.capabilities.maxImageCount;
	}
        // ...
```

Adding this feature via a PR is easy enough (I'd be happy to do it), but seeing how "confidently" the code chooses `minImageCount + 1`, I'm now wondering if my request makes sense in practice.

Out of curiosity, is it really so beneficial to favor a triple-buffering setup over double-buffering? (was this observed in practice?)
Is it a bad idea nowadays to prefer double buffering in order to spare some resources, but possibly at the expense of presentation rate?

Thanks!

charles-lunarg:
I think the original code should be amended. This logic was a bit ad hoc, taken from my own code and other only samples. The +1 is from such code.

Being able to *set* the image count directly is technically impossible, as the number is decided by the presentation engine ultimately. All we can do is set the absolute min and max and the presentation engine will choose something in between.

yoanlcq:
Alright, thanks! Then I'll submit a PR for that some time later today, if that's OK.

I think the API should just provide `set_desired_min_image_count()`, and the value is clamped when building; otherwise, if that is not used, then keep the default behavior of `minCount+1`.
The most flexible way would be to provide a function pointer to select the image count from the min and max, but that looks too overkill and niche to be worth it.

charles-lunarg:
Closed as resolved.

PR #138, opened 2022-05-16
SwapchainBuilder: provide some control over minImageCount

yoanlcq:
Hi,

This is a suggested fix for [#137](https://github.com/charles-lunarg/vk-bootstrap/issues/137).

I'm open to any feedback.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1983,7 +1983,7 @@ SwapchainBuilder& SwapchainBuilder::set_desired_min_image_count(uint32_t min_ima
 	info.add_to_min_image_count = 0; // Unnecessary, but let's do it for clarity
 	return *this;
 }
-SwapchainBuilder& SwapchainBuilder::use_default_min_image_count(uint32_t add_to_min_image_count = 1) {
+SwapchainBuilder& SwapchainBuilder::use_default_min_image_count(uint32_t add_to_min_image_count) {
 	info.min_image_count = 0;
 	info.add_to_min_image_count = add_to_min_image_count;
 	return *this;
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1774,14 +1774,15 @@ detail::Result<Swapchain> SwapchainBuilder::build() const {
 		return detail::Error{ SwapchainError::failed_query_surface_support_details, surface_support_ret.vk_result() };
 	auto surface_support = surface_support_ret.value();
 
-	uint32_t image_count;
+	uint32_t image_count = info.min_image_count;
 	if (info.required_min_image_count >= 1) {
 		if (info.required_min_image_count < surface_support.capabilities.minImageCount)
 			return make_error_code(SwapchainError::required_min_image_count_too_low);
 		
 		image_count = info.required_min_image_count;
 	} else if (info.min_image_count == 0) {
-		image_count = surface_support.capabilities.minImageCount + 1; // This has been the default behavior so far.
+		// We intentionally use minImageCount + 1 to maintain existing behavior, even if it typically results in triple buffering on most systems. 
+		image_count = surface_support.capabilities.minImageCount + 1;
 	} else {
 		image_count = info.min_image_count;
 		if (image_count < surface_support.capabilities.minImageCount)
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1774,10 +1774,18 @@ detail::Result<Swapchain> SwapchainBuilder::build() const {
 		return detail::Error{ SwapchainError::failed_query_surface_support_details, surface_support_ret.vk_result() };
 	auto surface_support = surface_support_ret.value();
 
-	uint32_t image_count = surface_support.capabilities.minImageCount + 1;
+	uint32_t image_count;
+	if (info.min_image_count == 0) {
+		image_count = surface_support.capabilities.minImageCount + info.add_to_min_image_count;
+	} else {
+		image_count = info.min_image_count;
+		if (image_count < surface_support.capabilities.minImageCount)
+			image_count = surface_support.capabilities.minImageCount;
+	}
 	if (surface_support.capabilities.maxImageCount > 0 && image_count > surface_support.capabilities.maxImageCount) {
 		image_count = surface_support.capabilities.maxImageCount;
 	}
+
 	VkSurfaceFormatKHR surface_format =
 	    detail::find_surface_format(info.physical_device, surface_support.formats, desired_formats, info.format_feature_flags);
 
@@ -1968,6 +1976,16 @@ SwapchainBuilder& SwapchainBuilder::use_default_format_feature_flags() {
 }
 SwapchainBuilder& SwapchainBuilder::set_image_array_layer_count(uint32_t array_layer_count) {
 	info.array_layer_count = array_layer_count;
+	return *this;
+}
+SwapchainBuilder& SwapchainBuilder::set_desired_min_image_count(uint32_t min_image_count) {
+	info.min_image_count = min_image_count;
+	info.add_to_min_image_count = 0; // Unnecessary, but let's do it for clarity
+	return *this;
+}
+SwapchainBuilder& SwapchainBuilder::use_default_min_image_count(uint32_t add_to_min_image_count = 1) {
+	info.min_image_count = 0;
+	info.add_to_min_image_count = add_to_min_image_count;
 	return *this;
 }
 SwapchainBuilder& SwapchainBuilder::set_clipped(bool clipped) {
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1776,7 +1776,7 @@ detail::Result<Swapchain> SwapchainBuilder::build() const {
 
 	uint32_t image_count;
 	if (info.min_image_count == 0) {
-		image_count = surface_support.capabilities.minImageCount + info.add_to_min_image_count;
+		image_count = surface_support.capabilities.minImageCount + 1; // This has been the default behavior so far.
 	} else {
 		image_count = info.min_image_count;
 		if (image_count < surface_support.capabilities.minImageCount)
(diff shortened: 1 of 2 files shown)


src/VkBootstrap.h
```diff
@@ -844,6 +844,13 @@ class SwapchainBuilder {
 	// Set the number of views in for multiview/stereo surface
 	SwapchainBuilder& set_image_array_layer_count(uint32_t array_layer_count);
 
+	// Sets the desired minimum image count for the swapchain. Note that the presentation engine is always free to create more images than requested.
+	SwapchainBuilder& set_desired_min_image_count(uint32_t min_image_count);
+	// Use the default desired minimum image count for the swapchain. The chosen value is calculated as `capabilitites.minImageCount + add_to_min_image_count`.
+	// The default behavior so far was equivalent to setting `add_to_min_image_count` to 1, and generally led to a triple buffering setup.
+	// A discussion of the trade-offs involved between double buffering and triple buffering is available as https://en.wikipedia.org/wiki/Multiple_buffering
```
charles-lunarg, line 851: Not sure this is the place to link to a 'general' discussion of multiple buffering, especially since this is in the context of vulkan. If there was a more vulkan focused link I'd be all for it.

src/VkBootstrap.cpp
```diff
@@ -1970,6 +1978,16 @@ SwapchainBuilder& SwapchainBuilder::set_image_array_layer_count(uint32_t array_l
 	info.array_layer_count = array_layer_count;
 	return *this;
 }
+SwapchainBuilder& SwapchainBuilder::set_desired_min_image_count(uint32_t min_image_count) {
+	info.min_image_count = min_image_count;
+	info.add_to_min_image_count = 0; // Unnecessary, but let's do it for clarity
+	return *this;
+}
```
charles-lunarg, line 1985: I think this interface would be a lot cleaner if instead of specifying the exact image count, there was an enum that controlled whether to double or triple buffer. 
Is it realistic that there is a need to control anything beyond setting it to one or two?
The presentation engine is pretty free to 'choose' any number higher that minImageCount after all.

src/VkBootstrap.h
yoanlcq: Yes, I wasn't too sure about it either; I haven't searched for Vulkan-specific resources because the concepts apply to any API.
I don't mind removing it.

src/VkBootstrap.cpp
yoanlcq: If this was an enum, it would be a 1-to-1 mapping to an image count, and there is no risk that the mapping may change, so at that point one might as well use the integer directly.
To me, it is as if we defined an enum such as "INT_0", "INT_1" and so on for each integer, just to force the use of a name rather than an integer which, at the end, is the value actually used.

I agree that enums are more type-safe and improve readability in general, however in this particular case I find it less fitting : 
- "min_image_count = 2" reads the same to me as "desire a double buffer";
- Plain integers provide a strict superset of the functionality of the enum; it's very likely that few people want to go beyond triple buffering, but we also have to reason to introduce an artificial limitation that prevents them from doing so.

In any case, the real thing that happens at the end is that we set `create_info.minImageCount`, and I would expect the API to match this closely.
I think `set_desired_min_image_count(N)` tells the reader evertyhing they need to know already: you'd like at least N images, but the presentation engine may provide more. I think an API such as `set_desired_multiple_buffering(DOUBLE_BUFFER)` is at best equivalent, but at worst brings less value.

That's my opinion at least.

charles-lunarg: I really would prefer an enum here for several reasons:

1. Much clearer to everyone. Double and Triple buffering are established terms. `set_swapchain_buffer_mode(DOUBLE_BUFFER)` is much more immediately obvious than `set_desired_min_image_count()`. Part of vk-bootstraps design was to make it "easier" for people new to the API to get up and running. Its a fine line using the correct terminology vs using an abstracted form that signals intent better. In other words, I disagree that `set_desired_min_image_count(N)` has enough information for a user to intuit what it does.

2. I am unaware of any practical use cases for setting it above 3. This is the first issue for vk-bootstrap discussing this count, which I interpret as most people not being that particular (and that triple buffering is a good default).

3. There is an information gap. If users want to reliably set the minImageCount to be triple or double buffered, they first need to know whether the presentation engine supports it. Yet I don't believe there is a mechanism for that currently in vk-bootstrap. `VkSurfaceCapabilitiesKHR.minImageCount` only requires that the value be 1 or greater. So to know what value to use, you'd need to first know what the minImageCount is. This isn't an issue with the direct value setting with `set_desired_min_image_count()` but is a problem with `use_default_min_image_count(add_to_min_image_count)` when `add_to_min_image_count` isn't zero. 

There is a design issue of whether to clamp or return an error if `set_desired_min_image_count()` can't be satisfied. Should vk-bootstrap silently triple buffer when the user asked for double buffering? (Yes, this applies to the enum solution, so it needs to be decided in either case). I lean towards just clamping it as its better to have a working swapchain than one which may not be 'perfect'. If doing exactly what they want is desired, they wouldn't likely be using vk-bootstrap.

nickclark2016: Why not have a `set_desired_min_image_count` and a `set_required_min_image_count` API? If a required min image count is provided and the surface doesn't satisfy that constraint, it should return an error.  If a required min image count isn't set and a desired is and the surface doesn't satisfy the desired constraint, clamp and return a swapchain.

As for DOUBLE_BUFFER and TRIPLE_BUFFER, how about providing named constants that are just integer values for 2 and 3?

charles-lunarg:
Okay after further reading and discussion, I acquiesce on there *needing* to be a boolean. I just want the function name/comment to be clear that its for controlling double and triple buffering. Being able to set the value directly isn't a "so dumb as to be impossible to use wrong" api, but the rest of the swapchain builder lets the user set the values directly regardless. So... it would be inconsistent to require an enum here but let users provide everything else as-is.

But the `use_default_min_image_count(uint32_t add_to_min_image_count)` needs to go. Being able to set the default min is fine, but the `add_to_min_image_count` has dubious usefulness and muddies the functions purpose. Is it for resetting the min_image_count? or adding to it?

yoanlcq:
> But the use_default_min_image_count(uint32_t add_to_min_image_count) needs to go

I'm fine with that, to be honest I followed the pattern for some of the other setters where you could "go back" to the default, and the parameter defaulting to 1 would lead to the equivalent of the behavior prior to this PR; but I agree with your comment on it and I'll do something about it.

> I do like the idea of adding const int DOUBLE_BUFFERING = 2; const int TRIPLE_BUFFERING = 3;. Very convenient way of being explicit without restricting the user.

I'm not against it; I would personnally never use such constants, but I understand the appeal, especially for newer users, tutorial writers, samples, or anyone looking for such symbols in a large codebase.

I've also never seen any use case for 4+ buffering... It's just that one of my principles when designing APIs is "if it's possible and it doesn't bother the common case, then provide it, even if you think nobody will use it", or simpler, "stay close to the truth and inform users so they can make their decisions".

Years ago, people used single buffering; then it quickly became double-buffering; then in recent years we've started to favor triple-buffering. The only thing I know is that I cannot predict if this will keep increasing.  
In doubt, just provide the ability to do it; it doesn't add complexity, and regardless of what we do, motivated enough people could bypass `vk-bootstrap` just to do it. So I figure, there would be no gain from preventing such uses.

yoanlcq:
One question remains though, should we actually preserve the existing behavior (which selects 2 or 3 depending on the driver) ?
I was a priori in favor of keeping it, to prevent unexpected "breakage", but since it was unpredictable anyway I'm not so sure anymore.

src/VkBootstrap.cpp
```diff
@@ -1774,10 +1774,23 @@ detail::Result<Swapchain> SwapchainBuilder::build() const {
 		return detail::Error{ SwapchainError::failed_query_surface_support_details, surface_support_ret.vk_result() };
 	auto surface_support = surface_support_ret.value();
 
-	uint32_t image_count = surface_support.capabilities.minImageCount + 1;
+	uint32_t image_count;
```
charles-lunarg, line 1777: I generally prefer defaulting all variables unless absolutely necessary that they be uninitialized.
```suggestion
	uint32_t image_count = info.min_image_count;
```

src/VkBootstrap.cpp
```diff
@@ -1774,10 +1774,23 @@ detail::Result<Swapchain> SwapchainBuilder::build() const {
...
+	if (info.required_min_image_count >= 1) {
+		if (info.required_min_image_count < surface_support.capabilities.minImageCount)
+			return make_error_code(SwapchainError::required_min_image_count_too_low);
+		
+		image_count = info.required_min_image_count;
+	} else if (info.min_image_count == 0) {
+		image_count = surface_support.capabilities.minImageCount + 1; // This has been the default behavior so far.
```
charles-lunarg, line 1784:
```suggestion
                // We intentionally use minImageCount + 1 to maintain existing behavior, even if it typically results in triple buffering on most systems. 
		image_count = surface_support.capabilities.minImageCount + 1;
```

src/VkBootstrap.h
```diff
@@ -844,6 +845,27 @@ class SwapchainBuilder {
 	// Set the number of views in for multiview/stereo surface
 	SwapchainBuilder& set_image_array_layer_count(uint32_t array_layer_count);
 
+	// Convenient named constants for passing to set_desired_min_image_count().
+	// Note that it is not an `enum class`, so its constants can be passed as an integer value without casting
+	// In other words, these might as well be `static const int`, but they benefit from being grouped together this way.
```
charles-lunarg, line 850: Well, this answers why use enum and not enum class. Though, with that comment I kinda wonder if they should be static const int, so that any code cleanliness warnings don't get issued for not using enum class.

Your call, I'm fine with either way.

charles-lunarg (commented):
Changes look good for what was discussed, through I have some nitpick's.

charles-lunarg (approved):

charles-lunarg:
Thanks!

yoanlcq:
About the enum class stuff I agree, but then if we make it enum class we have to add overloads for `set_desired_min_image_count`  and `set_required_min_image_count`, and in general any time we provide an API that accepts an image count.

That would be two more lines, but it's more the idea of duplicating functions that I'm not really a fan of.

An advantage of using an enum instead of static const int is also that debuggers will show the enum constant's name rather than just the integer. In practice it won't matter here, but it's one of the reasons I use enums in such situations in general.

Hence I'm leaving the enum as is, hoping it's OK.

PR #139, opened 2022-05-16
SwapchainBuilder: expose present_mode and requested_min_image_count in the resulting Swapchain

yoanlcq:
Hi,

It is as the title says; my use cases are : 

- I wanted a feature to get the present mode actually used for the swapchain, so I could display an indicator in my application;
- I wanted to easily get the value of `create_info.minImageCount`, to confirm (and display) whether or not we're likely in a double-buffering setup or triple-buffering setup. I know this does not prove anything, but I think it is interesting even as a hint.

Thanks for considering this, as always I am open to any feedback.


charles-lunarg (approved):
Changes look appropriate.

Would be nice to have a test which sets the present mode and verifies that it got it.

PR #146, opened 2022-06-09
Reverse alias registration for types

Eearslya:
Aliases are defined in `vk.xml` with a `name` and `alias` attribute. The desired outcome is for the symbol in `name` to point to the symbol in `alias`.

e.g.
```xml
<type category="struct" name="VkSemaphoreWaitInfoKHR" alias="VkSemaphoreWaitInfo"/>
```
is equivalent to:
```cpp
typedef VkSemaphoreWaitInfo VkSemaphoreWaitInfoKHR;
```

The dispatch generator script was doing the opposite, taking the "fake" aliased name and translating it back to the original. While this worked for most types, `VkPipelineInfoEXT` is a struct that was only added to the headers [in revision 213](https://github.com/KhronosGroup/Vulkan-Headers/blame/ec9b96aae53e152f6812e66fa43757c1256369a0/include/vulkan/vulkan_core.h#L13885), while the current SDK is only 211.

While this issue would disappear with SDK 213 or above, I believe this change is the way aliases are intended to be used.

This change will adjust all of the function prototypes in vk-bootstrap to use the core version of the struct, if available. Common best practices say you should use the core version anyhow, so this could be seen as an improvement. More importantly, it now compiles with the latest SDK thanks to the removal of `VkPipelineInfoEXT`.


charles-lunarg:
The use of the extension type was actually intentional because it allows using vkbootstrap with headers older than the one it was compiled with. However, this is a case of a newer extension breaking the assumption that EXT types are perfect aliases to KHR types. 

A solution would be to special case VkPipelineInfoEXT/KHR by creating a list of aliases that need to be manually 'corrected'.

Except that half of the code changed here doesn't use the 'right' type to begin with. 
For instance `vkCmdBeginRenderPass2` shouldn't need the `VkSubpassBeginInfoKHR`, its a core function. Only `vkCmdBeginRenderPass2KHR` should need the `VkSubpassBeginInfoKHR` alias. This PR reverses the aliasing, which fixes the logic in some places but doesn't fix others, such as for `vkCmdBeginRenderPass2KHR`.

In short: This is a hard problem and the spec makes it harder by having aliases which sometimes are fine to use and other times aren't.

Eearslya:
You make a good point, though I would say that developers should always have the latest SDK, since every change is backwards-compatible anyway.

I wouldn't say this is an extension "breaking EXT to KHR aliases". If anything, the issue here is that vk-bootstrap is generating header files based on a newer version of the Vulkan headers than the SDK currently provides.

As of header 213, the EXT variant *is* a perfect alias of the KHR variant. The issue is simply that vk-bootstrap knows about it and those on header 211 don't yet.

charles-lunarg:
It really comes back to vk-bootstrap being generated with newer headers than in the SDK. 

Would it make sense to create a 'dev' branch that newer header versions are added to, and so the latest version on master will be for the latest released SDK rather than headers newer than that?

(also kinda a solution to the opened issue, though not a stellar one)

Eearslya:
Seems like that might be the best solution, unless there is some programmatic way to determine what the latest SDK is. Scraping the download page would be a fragile nightmare.

It would just require someone to push the big 'merge' button every time a new SDK comes out.

charles-lunarg:
Its fine to only push out to master once the SDK is released, but I don't enjoy being behind the latest headers for the branch everyone pulls from. Not that it would be a big deal, but it would mean remembering to update this repo after an SDK is released.

It makes me wonder if there is a better way to guard the necessary contents better. I should probably add a CI check to make sure the code is compiled against older headers, to catch this kind of regression.

charles-lunarg:
I'm going to close this for now as I have merged a fix for the issue with VkPipelineInfoEXT/KHR. While I appreciate the spirit of this PR, it does necessarily make everyone have to use the newest header version to use newer functionality. Not that the current implementation does a perfect job of allowing older header usage anyhow...

Feel free to reopen if you feel that this PR still has merit, or make an issue to further more general discussion to the issue.

PR #154, opened 2022-07-08
Fix crash during removal of unsuitable physical devices when using only_fully_suitable selection mode

Valakor:
The second argument to erase is exclusive, so we want to use `.end()` directly; otherwise we'll either fail to remove the very last element (if the range is non-empty) or crash (if `partition_index` is `physical_devices.end()`).


charles-lunarg (approved):
Whoops! Seems I didn’t read the docs thoroughly enough. Thanks for the fix!

PR #158, opened 2022-07-22
Expose the final color space used when creating swapchains

Valakor:
This just passes through the final color space used after finding a suitable surface format to the user so it can be queried later.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1852,6 +1852,7 @@ detail::Result<Swapchain> SwapchainBuilder::build() const {
 	}
 	swapchain.device = info.device;
 	swapchain.image_format = surface_format.format;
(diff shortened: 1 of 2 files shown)


charles-lunarg (approved):
Other than the extraneous commits in the PR, these changes look fine to me.

Valakor:
Whoops, sorry about that. Those other commits have already been merged and shouldn't technically be part of this PR. Only the most recent 2 commits matter here.

My git-foo isn't good enough to know exactly how to go about rebasing things against this repo to only include the two commits I care about.

charles-lunarg:
What I would do is this:
`git rebase -i HEAD~5` to get into interactive rebase mode
Mark all the commits you do not want as `drop` or `d`, then save.
Unless something terrible happened, that should be enough to get rid of the extra commits.
Then just force push to get the changes up to github.

Issue #169, opened 2022-10-03
SwapchainBuilder format_feature_flags default doesn't seem appropriate


ovatonne:
The default format feature flag in SwapchainBuilder specified in this line doesn't make sense to me.
https://github.com/charles-lunarg/vk-bootstrap/blob/5b15f6e881f354c71bca478ea99e4906da32dfc3/src/VkBootstrap.h#L844-L846

Because the default image usage flag is VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, the default format feature should be VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT instead of VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT. 
Also, I think for bootstraping Vulkan applications, swapchain readback by sampling the image is not the very first use-case one wants to consider.
By default, this prevents any format that is "write/transfer-only" to be used. I think this never happens on any implementation (usually formats can be sampled but not written to optimally), but still this default doesn't seem right.

Furthermore, there are no checks done during the selection to verify that the requested image usage is possible with the format. This should be caught by validation layers at [VUID-VkSwapchainCreateInfoKHR-imageFormat-01778](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSwapchainCreateInfoKHR.html) (sorry, I don't know how to link to the correct line), but because most of this library checks valid usages, this verification seems missed.

If I understand properly I think the functions to set the desired VkFormatFeatureFlags shouldn't be used. The VkFormatFeatureFlags should only be a mean to verify that the VkImageUsageFlags set by `set_image_usage_flags` are supported by the desired formats.

Disclaimer: This is the first time I create an issue on a Vulkan github. Sorry if I am missing something in my understanding or explanation.

If any of the above makes sense, a plan of action to change the behavior would be from less to more breaking:

1. Deprecating the use of the requested format feature flag
2. Changing the default
3. Not using the provided format feature flag
4. Checking  [VUID-VkSwapchainCreateInfoKHR-imageFormat-01778](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSwapchainCreateInfoKHR.html) during the format selection, and return an error no format supports the provided image usage.

I would be happy to make a MR if you think some of the proposed changes are reasonable.

charles-lunarg:
Yes, this looks like a wrong default. That definitely should be changed.

[...]

PR #173, opened 2022-10-10
Stricter verifications in SwapchainBuilder and other default format feature

ovatonne:
To solve #169, I chose the following approach:

1. Remove default format_feature_flag (set to 0)
2. Enforce checking of user requirements (format features, image usage and desired formats)
3. To check image usage, only basic verification is done, by checking that the usage match the swapchain capabilities (in the usual case of non-shared present modes)

See below some notes on the change to explain the decision:

1. Even _VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT_ is not a good default because it implies that the format should support usage _VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT_, which is not mandatory by the spec (color attachment is the only mandatory usage for implementers).
2. This broke a test on my machine because it doesn't support _VK_FORMAT_R8G8B8A8_UNORM_ as swapchain format. It should have never passed though - maybe validation caught the issue, but I couldn't see it.
--> This may be a breaking change
3. I am not sure this is enough validation, but it should be a good start.

Don't hesitate changing the code, or commenting on this. I am motivated to improve this if needed !

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -443,7 +443,7 @@ const char* to_string(SwapchainError err) {
 		CASE_TO_STRING(SwapchainError, failed_get_swapchain_images)
 		CASE_TO_STRING(SwapchainError, failed_create_swapchain_image_views)
 		CASE_TO_STRING(SwapchainError, required_min_image_count_too_low)
-		CASE_TO_STRING(SwapchainError, required_usage_or_features_not_supported)
+		CASE_TO_STRING(SwapchainError, required_usage_not_supported)
 		default:
 			return "";
 	}
@@ -1585,8 +1585,7 @@ enum class SurfaceSupportError {
 	surface_handle_null,
 	failed_get_surface_capabilities,
 	failed_enumerate_surface_formats,
-	failed_enumerate_present_modes,
-	no_suitable_format
+	failed_enumerate_present_modes
 };
 
 struct SurfaceSupportErrorCategory : std::error_category {
@@ -1633,7 +1632,7 @@ Result<SurfaceSupportDetails> query_surface_support_details(VkPhysicalDevice phy
 	return SurfaceSupportDetails{ capabilities, formats, present_modes };
 }
 
-Result<VkSurfaceFormatKHR> find_surface_format(VkPhysicalDevice phys_device,
+VkSurfaceFormatKHR find_surface_format(VkPhysicalDevice phys_device,
     std::vector<VkSurfaceFormatKHR> const& available_formats,
     std::vector<VkSurfaceFormatKHR> const& desired_formats,
     VkFormatFeatureFlags feature_flags) {
@@ -1648,8 +1647,8 @@ Result<VkSurfaceFormatKHR> find_surface_format(VkPhysicalDevice phys_device,
 		}
 	}
 
-	// If no format supports the provided feature, we report that no format is suitable to the user request
-	return { make_error_code(SurfaceSupportError::no_suitable_format) };
+	// use the first available one if any desired formats aren't found
+	return available_formats[0];
 }
 
 VkPresentModeKHR find_present_mode(std::vector<VkPresentModeKHR> const& available_resent_modes,
@@ -1772,10 +1771,8 @@ Result<Swapchain> SwapchainBuilder::build() const {
 		image_count = surface_support.capabilities.maxImageCount;
 	}
 
-	auto surface_format_ret =
+	VkSurfaceFormatKHR surface_format =
 	    detail::find_surface_format(info.physical_device, surface_support.formats, desired_formats, info.format_feature_flags);
-	if (!surface_format_ret.has_value()) return Error{ SwapchainError::required_usage_or_features_not_supported };
-	auto surface_format = surface_format_ret.value();
 
 	VkExtent2D extent = detail::find_extent(surface_support.capabilities, info.desired_width, info.desired_height);
 
@@ -1791,7 +1788,7 @@ Result<Swapchain> SwapchainBuilder::build() const {
 
 	if (detail::is_unextended_present_mode(present_mode) &&
 	    (info.image_usage_flags & surface_support.capabilities.supportedUsageFlags) != info.image_usage_flags) {
-		return Error{ SwapchainError::required_usage_or_features_not_supported };
+		return Error{ SwapchainError::required_usage_not_supported };
 	}
 
 	VkSurfaceTransformFlagBitsKHR pre_transform = info.pre_transform;
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1633,7 +1633,7 @@ Result<SurfaceSupportDetails> query_surface_support_details(VkPhysicalDevice phy
 	return SurfaceSupportDetails{ capabilities, formats, present_modes };
 }
 
-Result<VkSurfaceFormatKHR> find_desired_surface_format(VkPhysicalDevice phys_device,
+Result<VkSurfaceFormatKHR> find_desired_surface_format(VkPhysicalDevice /* phys_device */,
     std::vector<VkSurfaceFormatKHR> const& available_formats,
     std::vector<VkSurfaceFormatKHR> const& desired_formats) {
 	for (auto const& desired_format : desired_formats) {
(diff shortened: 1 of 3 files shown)


src/VkBootstrap.cpp
```diff
@@ -1648,8 +1648,8 @@ VkSurfaceFormatKHR find_surface_format(VkPhysicalDevice phys_device,
 		}
 	}
 
-	// use the first available one if any desired formats aren't found
-	return available_formats[0];
```
charles-lunarg, line 1652: I have a feeling that this will result in breaking user code. The fact that the test code had to be modified by adding VK_FORMAT_B8G8R8A8_UNORM makes me leery about this change. Now, that isn't to say the change to return an error instead of the first format isn't good, as it definitely is more consistent & logical. The issue is making sure that the 'person who just wants a swapchain with whatever format is available' is still supported.

src/VkBootstrap.cpp
```diff
@@ -1664,6 +1664,11 @@ VkPresentModeKHR find_present_mode(std::vector<VkPresentModeKHR> const& availabl
 	return VK_PRESENT_MODE_FIFO_KHR;
 }
 
+bool is_unextended_present_mode(VkPresentModeKHR present_mode) {
+	return (present_mode == VK_PRESENT_MODE_IMMEDIATE_KHR) || (present_mode == VK_PRESENT_MODE_MAILBOX_KHR) ||
+	       (present_mode == VK_PRESENT_MODE_FIFO_KHR) || (present_mode == VK_PRESENT_MODE_FIFO_RELAXED_KHR);
+}
```
charles-lunarg, line 1670: Why do we need this check here? Does the present mode not being one of the 4 Vulkan 1.0 present modes pose additional usage/format feature requirements? If so, that should be documented in the implementation code that uses this function.

src/VkBootstrap.h
```diff
@@ -849,11 +851,14 @@ class SwapchainBuilder {
...
+	// If the desired formats cannot allow it, building the swapchain will result in the
+	// `SwapchainError::required_usage_or_features_not_supported` error. Use this functionnality only if you require
+	// formats to support a set of feature flags that are not already implied by set_image_usage_flags
 	SwapchainBuilder& set_format_feature_flags(VkFormatFeatureFlags feature_flags);
 	// Add a format feature to the bitmask for acquired swapchain images.
 	SwapchainBuilder& add_format_feature_flags(VkFormatFeatureFlags feature_flags);
 	// Use the default format feature bitmask values. This is the default if no format features
-	// are provided. The default is VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT
+	// are provided. The default is 0
 	SwapchainBuilder& use_default_format_feature_flags();
```
charles-lunarg, line 862: Why keep the ability to set format feature flags? My understanding is that format feature flags are something the application queries from the driver about a format, and the usage flags are used to tell the driver what they want to do with an image. To me, that would mean we only need to keep usage flags around for the user to do. Definitely let me know if I am mistaken about this.

src/VkBootstrap.cpp
ovatonne: Ok, I have definitely misunderstood the intent of the "desired" formats. I though here it was returning the first format the user desired, but it is just one format a user requested. It explains a lot why the test was the way it was :)
I will definitely revert this, as changing this behavior was definitely not correct !

src/VkBootstrap.cpp
ovatonne: It is because of the Note explanation on [VkSurfaceCapabilitiesKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSurfaceCapabilitiesKHR.html) stating that the `supportedUsageFlags` only applies to those present modes, and not the shared ones. One would have to check if `VkSharedPresentSurfaceCapabilitiesKHR ` extension is enabled to perform further validation, but it goes out of scope of the work I wanted to do for this PR. I felt it would be a nice "simple" addition, but I didn't want to do the heavier work the validation layers do.

On next commit, I will add comments on this function so that it is clear. I could remove this validation because the layers do it anyway, but I felt that it would improve the library at small cost.

src/VkBootstrap.h
ovatonne: First, I backtracked a bit because I didn't want to remove a function that some people may use.
Then, I wasn't sure if anyone relied on this (if there is any use-case where one has to use format features instead of image usages). 

My understanding is the same as yours. Furthermore, the mapping `image usage` --> `format feature flags` seems to be 1 --> N, meaning that if the implementation reports a supported usage in `vkGetPhysicalDeviceSurfaceCapabilitiesKHR`, it also supports all associated format features. Thus it would be irrelevant to use the format feature flags

So for the very unique case where the driver reports `supportedUsageFlags` in `vkGetPhysicalDeviceSurfaceCapabilitiesKHR`, instead of the usual feature flags, we shouldn't need to query feature flags anymore. It is still unclear to me why they report supportedUsages instead of FeatureFlags.

Since Monday, I convinced myself that this API is not useful, so I will remove it in next commit.

ovatonne:
I have made the changes stated in the comments. Now it really looks like what I had in mind when I wrote Issue https://github.com/charles-lunarg/vk-bootstrap/issues/169

src/VkBootstrap.cpp
```diff
@@ -1633,22 +1633,29 @@ Result<SurfaceSupportDetails> query_surface_support_details(VkPhysicalDevice phy
 	return SurfaceSupportDetails{ capabilities, formats, present_modes };
 }
 
-VkSurfaceFormatKHR find_surface_format(VkPhysicalDevice phys_device,
+Result<VkSurfaceFormatKHR> find_desired_surface_format(VkPhysicalDevice /* phys_device */,
     std::vector<VkSurfaceFormatKHR> const& available_formats,
```
charles-lunarg, line 1637: Should remove the phys_device parameter if it is unused.

src/VkBootstrap.cpp
```diff
@@ -1633,22 +1633,29 @@ Result<SurfaceSupportDetails> query_surface_support_details(VkPhysicalDevice phy
...
 		}
 	}
 
-	// use the first available one if any desired formats aren't found
+	// if no desired format is available, we report that no format is suitable to the user request
+	return { make_error_code(SurfaceSupportError::no_suitable_desired_format) };
+}
+
+VkSurfaceFormatKHR find_best_surface_format(VkPhysicalDevice phys_device,
+    std::vector<VkSurfaceFormatKHR> const& available_formats,
```
charles-lunarg, line 1653: Same as above, if no phys dev is used, remove the parameter.

charles-lunarg (commented):
Some nit picks.

src/VkBootstrap.cpp
charles-lunarg: Well, now that I look at the code, I realize that there isn't a way to specify 'required' formats, which seems like a hole. 

The original design is a 'use the first format the driver gives us *unless* the user asks for something else and it is supported. 
It probably should be amended to allow specifying a list 'required' formats, where the first supported format in the list is chosen, but if no formats are chosen, and error is returned. This could be relatively seamlessly added by having an error be returned if the required list is not empty. So the existing behavior is maintained (if you specify nothing you get the first format), if you specify a desired format, you get it if supported, otherwise the first in the list, and if you speciy a required format, you get it if supported or an error.

Don't feel like you need to implement this, I'm more spitballing ideas than listing asking for changes.

src/VkBootstrap.cpp
charles-lunarg: That makes sense. Thanks for the clarification (and yeah, its kinda annoying how there are present modes for desktop and present modes for mobile, with different requirements on each...

src/VkBootstrap.h
charles-lunarg: Great, the removal makes the interface just that bit smaller and thus simpler.

Plus, while vk-bootstrap tries to validate what it can, it can only do so much within reason. Better to let validation layers do the heavy lifting.

charles-lunarg:
When the nits are resolved, I'll approve as this is in a good shape.

src/VkBootstrap.cpp
ovatonne: Yes, I though at first "desired" meant "required" for some reason, hence my first change.
I specifically left the two separate functions for simple refactoring later (reusing the first function with the error code should be sufficient to check required formats).
Another PR will be best suited for this, but it is certainly a good feature to add !

src/VkBootstrap.cpp
ovatonne: Should be resolved. Same as below.

charles-lunarg (approved):

PR #181, opened 2022-10-28
:: instead of . required

brgalo:
my first issue, just ran into this while trying it out myself

Is this the correct way to do this?


charles-lunarg (approved):
Good catch, and yes this is how to do it.

PR #187, opened 2022-12-18
Fix graphics/present queue assignment

rogmatic:
Fixed SwapchainBuilder second constructor, which had 'info.graphics_queue_index = present.value()' and 'info.present_queue_index = graphics.value()'.


charles-lunarg (approved):
Whoops! Very simple mistake, thanks for catching and fixing it :)

(I was Out of office for a while so thats why it took some time to respond)

Issue #165, opened 2022-08-30
Allow to disable tests if built as a standalone


SpaceIm:
Tests & examples can't be disabled if vk-bootstrap is built from top CMakeLists directly.

Please replace by this logic for example:

```cmake
if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
    set(VK_BOOTSTRAP_TEST_DEFAULT_VALUE ON)
else()
    set(VK_BOOTSTRAP_TEST_DEFAULT_VALUE OFF)
endif()
option(VK_BOOTSTRAP_TEST "Test Vk-Bootstrap with glfw and Catch2" ${VK_BOOTSTRAP_TEST_DEFAULT_VALUE})

if(VK_BOOTSTRAP_TEST)
    add_subdirectory(ext)
    add_subdirectory(tests)
    add_subdirectory(example)
endif ()
```

(but I would advice to always disable `VK_BOOTSTRAP_TEST` by default, and enable it in your CI files)

charles-lunarg:
@SpaceIm Done

Thats definitely an oversight on my part, thanks for catching it. And for providing a *very* clean solution.

PR #194, opened 2023-02-15
Added cmake options to hide warnings if is not main project

Camba3D:
Similar to this: https://github.com/charles-lunarg/vk-bootstrap/issues/165
But to avoid warnings if vk-bootstrap is built from top CMakeLists directly.

- [x] Added cmake option VK_BOOTSTRAP_DISABLE_WARNINGS (OFF by default to preserve current behaviour)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 project(VulkanBootstrap)
 
+option(VK_BOOTSTRAP_DISABLE_WARNINGS "Enable warnings as errors during compilation" OFF)
 option(VK_BOOTSTRAP_WERROR "Enable warnings as errors during compilation" OFF)
 
 add_library(vk-bootstrap-vulkan-headers INTERFACE)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 project(VulkanBootstrap)
 
+option(VK_BOOTSTRAP_DISABLE_WARNINGS "Enable warnings as errors during compilation" OFF)
```
charles-lunarg, line 4: Possibly copy/paste mistake? (the description is the same as the other option)

Camba3D: Sure! Sorry for the copypaste haha. Updated!

charles-lunarg (approved):
Drat, you fixed the one thing I was going to comment on! /s

Looks good. I initially hesitated to accept this PR cause "more warning more gooder" buuuut as much as *I* should be enabling each and every warning and fixing them, users have no need to do that. Thus, making it cleaner to use is good.

PR #196, opened 2023-02-28
Add a method to disable the debug messenger

DethRaid:
Adds a method to completely disable the debug messenger. This will make vk-bootstrap not request the debug utils extension. Debug utils isn't available by default on platforms such as Android


src/VkBootstrap.cpp
```diff
@@ -794,6 +794,13 @@ InstanceBuilder& InstanceBuilder::request_validation_layers(bool enable_validati
 	info.request_validation_layers = enable_validation;
 	return *this;
 }
+
+InstanceBuilder & InstanceBuilder::disable_debug_messenger() {
+	info.use_debug_messenger = false;
+	info.debug_callback = nullptr;
```
charles-lunarg, line 800: Rather than setting the debug_callback to nullptr, I think it would make more sense to modify the code to not add the debug utils extension if use_debug_messenger is false.

The relevant code change is line 587 of VkBootstrap.c, listed below
```c
	if (info.debug_callback != nullptr && info.use_debug_messenger && system.debug_utils_available) {
		extensions.push_back(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
	}
```

charles-lunarg: Cause right now `info.use_debug_messenger` is only used to decide if the callback should be setup, not if the extension should be added.

DethRaid: I made the change, it works great for my Android

src/VkBootstrap.h
```diff
@@ -380,7 +380,7 @@ class InstanceBuilder {
 	InstanceBuilder& enable_validation_layers(bool require_validation = true);
 	// Checks if the validation layers are available and loads them if they are.
 	InstanceBuilder& request_validation_layers(bool enable_validation = true);
-
+
```
charles-lunarg, line 383: Nit: random whitespace :P

charles-lunarg (approved):
Great, this is super clean all in all.

PR #206, opened 2023-10-03
Fixed instance layer ordering causing validation layer issues with VK_LAYER_KHRONOS_shader_object

raff766:
Before this fix, enabling the `VK_LAYER_KHRONOS_shader_object` extension layer caused erroneous validation errors to pop up, seemingly as if the validation layers were not aware that the layer was enabled. Changing the order of the layers when building the instance fixed the issue.

For example, before this fix the validation layers would complain about the `VkStructureType` of `vk::PhysicalDeviceShaderObjectFeaturesEXT`  being a completely different type presumably because it doesn't recognize `VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_OBJECT_FEATURES_EXT`.

My code:
```cpp
    vkb::PhysicalDevice vkbPhysicalDevice = physicalDeviceSelector
            .set_minimum_version(1, 3)
            .add_required_extension(VK_EXT_SHADER_OBJECT_EXTENSION_NAME)
            .add_required_extension_features(vk::PhysicalDeviceShaderObjectFeaturesEXT { .shaderObject = true })
            .set_surface(rawSurface)
            .select()
            .value();
```
Error:
```
Validation Error: [ VUID-VkDeviceCreateInfo-pNext-pNext ] | MessageID = 0x901f59ec | vkCreateDevice: pCreateInfo->pNext
chain includes a structure with unexpected VkStructureType VK_STRUCTURE_TYPE_PRIVATE_DATA_SLOT_CREATE_INFO; Allowed stru
ctures are [VkDeviceDeviceMemoryReportCreateInfoEXT, VkDeviceDiagnosticsConfigCreateInfoNV, VkDeviceGroupDeviceCreateInf
o, VkDeviceMemoryOverallocationCreateInfoAMD, VkDevicePrivateDataCreateInfo, VkPhysicalDevice16BitStorageFeatures, VkPhy
sicalDevice4444FormatsFeaturesEXT, VkPhysicalDevice8BitStorageFeatures, VkPhysicalDeviceASTCDecodeFeaturesEXT...
]. This error is based on the Valid Usage documentation for version 2
61 of the Vulkan header.  It is possible that you are using a struct from a private extension or an extension that was a
dded to a later version of the Vulkan header, in which case the use of pCreateInfo->pNext is undefined and may not work
correctly with validation enabled The Vulkan spec states: Each pNext member of any structure (including this one) in the
 pNext chain must be either NULL or a pointer to a valid instance of VkDeviceDeviceMemoryReportCreateInfoEXT, VkDeviceDi
agnosticsConfigCreateInfoNV...
```

I haven't tested if similar problems occur with the other extension layers, but hopefully this will fix those too if that was the case.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -634,12 +634,12 @@ Result<Instance> InstanceBuilder::build() const {
         return make_error_code(InstanceError::requested_extensions_not_present);
     }
 
-    for (auto& layer : info.layers)
-        layers.push_back(layer);
-
     if (info.enable_validation_layers || (info.request_validation_layers && system.validation_layers_available)) {
         layers.push_back(detail::validation_layer_name);
     }
+    for (auto& layer : info.layers)
+        layers.push_back(layer);
+
     bool all_layers_supported = detail::check_layers_supported(system.available_layers, layers);
     if (!all_layers_supported) {
         return make_error_code(InstanceError::requested_layers_not_present);


charles-lunarg:
The listed validation message is a bug in the shader object layer. It indicates that a pNext chain contains a struct with an sType that shouldn't be there. [This has been fixed as of a month ago](https://github.com/KhronosGroup/Vulkan-ExtensionLayer/commit/53ab6ec5ef00a66d13e48efbdd936de713072de3). The fix will be in the next SDK's release of the shader object layer.

About this PR, I am generally against 'fixing' validation errors by simply hiding them. That is because what the change in this PR here does is cause the validation layer to only see calls sent by the application. Before, every call the shader object layer would make goes through validation first before hitting the driver. 

That doesn't mean this PR is *wrong*, just that I believe the fix isn't really solving the underlying problem.

I could see a change to allow choosing "where" the validation layer is in relation to other enabled layers, but that is a whole separate PR.

charles-lunarg:
Side note - you do not need to sign the google CLA to contribute to this repo. This isn't a google repo, but it appears when *I* contributed to a google repo recently, it added it to this one. I think I have fixed it, but I hope that didn't cause too much confusion.

raff766:
Ah I see, my fault for not testing with the latest extension layers. 🙂

> I could see a change to allow choosing "where" the validation layer is in relation to other enabled layers, but that is a whole separate PR.

Yeah that might be a good idea, it could help users debug/get around problems they're having with the layers such as in this case. Maybe making it so that where the call to `request_validation_layers()` or `enable_validation_layers()` is in relation to other `enable_layer()` calls determined the ordering of the layers could be a way to implement something like that. I could create a PR for that if you think that's a good idea.

But besides that, I think its safe to close out this PR then.

charles-lunarg:
> I could create a PR for that if you think that's a good idea.

If you created a PR for this, I'd be happy to review & accept. I personally think its not a priority so wouldn't want you to spend time on it, since I consider it low value. Whats fun is learning just how deep the 'layer ordering rabbit hole' goes (since I maintain the Vulkan-Loader, I have spent a significant amount of time finding out the weird and wild ways layers operate).

raff766:
Alright, I guess I'll leave it as is then, thanks for the insider info on the intricacies of Vulkan layers and how the ordering of the layers affects what calls the validation layers check 👍

PR #209, opened 2023-10-07
Fix API of CustomQueueDescription

nickclark2016:
Issue: `vkb::CustomQueueDescription` provides an interface that allows a mismatch between the size priorities vector and the count of queues.  This allows the user to provide a vector that is too small (and have a memory read out of bounds), or allows the user to provide a vector that is larger than the count and expect all elements in the vector to be used.

Solution: To solve this, we break the API.  Instead of consuming both a count and a vector, `vk-bootstrap` will use the size of the vector to determine queue count.  While it's not typically a good thing to break the API, this provides us a mechanism to make VKB safer with a trivial fix for code consuming this.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1437,10 +1437,8 @@ DispatchTable Device::make_table() const { return { device, fp_vkGetDeviceProcAd
 
 Device::operator VkDevice() const { return this->device; }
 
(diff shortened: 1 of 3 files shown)


charles-lunarg (approved):
Hate breaking API - but this is such a good cleanup that its worth doing. Especially since anyone *using* the CustomQueueDescription may have already realized that its duplicate information.

PR #214, opened 2023-10-17
Add new Queue query API

### LINKED ISSUE #197: [Feature Request] Get Queue and Queue Index Utility
A lot of times, it's nice to have both the VkQueue and queue index.  I propose we add this convenience function to `vkb::Device`:

```cpp
vkb::Result<std::pair<VkQueue, std::uint32_t>> vkb::Device::get_queue_and_index(vkb::QueueType type) const;
```

Upon failure to fetch the queue index or `VkQueue`, it would return a result with the respective error.  Upon successfully fetching both, it would return a pair of the queue and index.

charles-lunarg:
The original queue query API has shortcomings which cause confusion and make it harder to do the things people want with vk-bootstrap.

Design features of the new API:
* Choose a queue given a combination of queue types - Instead of only choosing based on a single queue type
* Allow picking the first or 'best' queue for any given combination of queue types
* Allow picking a present queue with a specific surface handle
	- Instead of only using the surface the PhysicalDevice was selected with
* Dont need to update the library to support new queue types over time

Fixes #197 
Replaces #175


charles-lunarg:
More details on why old API isn't great:

* Imagine hardware with 2 families, one that is graphics + compute + transfer, and one with compute + transfer. get_queue(transfer) returns the same queue handle as get_queue(compute), but the API makes it look like they are separate. And here is that exact situation on top of the line AMD hardware http://vulkan.gpuinfo.org/displayreport.php?id=25162#queuefamilies

* The custom QueueType enum made it appear as if each queue only supported 1 type of operations, hiding the complexity that actually exists. 

* Trying to make each queue type have its "own" queue is not practical given the wide variety of queue families 

The old API tried very hard to give 'unique' queues, `get_dedicated_queue()` for example, but the new API pushes that responsibility onto the user. There should be an easy way to determine uniqueness of queues. So.. the PR definitely isn't ready to merge.

charles-lunarg:
```c++
struct QueueDetails {
    VkQueueFlags desired_flags;
    bool found = false;
    VkQueue queue;
    uint32_t queue_family_index;
};

std::vector<QueueDetails> desired_queues = {
    { GRAPHICS | TRANSFER | COMPUTE },
    { COMPUTE },
    { TRANSFER },
};

vkb::Device.get_queues(desired_queues);

if (desired_queues[0].found) {
    uber_queue = desired_queues[0].queue;
}
if (desired_queues[1].found) {
    compute_queue = desired_queues[0].queue;
}
if (desired_queues[2].found) {
    transfer = desired_queues[0].queue;
}
```

Just thinking out loud, here is a potential API for getting queues from different families. 
The logic is that the app submits a list of requested queues by filling out the desired_flags member. Then, vk-bootstrap tries to find the best queue for each item, where each element that is satisfied is removed from considering for subsequent elements. 
Ie, every queue found is guaranteed to be unique, and non-unique queues are 'not found'.

PR #215, opened 2023-10-22
Add enable_extension_if_present to vkb::PhysicalDevice

charles-lunarg:
If the given extension is present, enable_extension_if_present will make the extension be enabled on the device.

This fixes a gap in capability due to the deprecation of add_desired_extension.


charles-lunarg:
@DexterHaxxor This is a quick solution to the deprecation of optional extensions. It allows for the easy enabling of extensions while also getting feedback if it is available or not.

PR #225, opened 2023-11-08
Unload Vulkan during vkDestroyInstance

charles-lunarg:
This commit adds a counter for the number of times the vulkan library is loaded. When this count reaches zero, it calls unload.


charles-lunarg:
Closing in favor of #226 
This PR tries to make it possible to unload vulkan 'automatically' with vkDestroyInstance, but the use case for this is better served by fixing some linking oddities due to nested classes & static functions/variables that prevented any shared library that statically linked vk-bootstrap from being unloaded properly in the first place.

PR #226, opened 2023-11-13
Cleanup internal class VulkanFunctions

charles-lunarg:
Removes the subclassing of VulkanLibrary, bringing its logic into the VulkanFunctions class. This resolves a weird linking issue which was causing hot-reloading of shared libraries that statically linked vk-bootstrap to fail to unload correctly, preventing reloading.
The commit also consolidates the logic in load_vulkan_funcs() into a single function rather than having it be split in three, which only made the code harder to reason about.

Also made vk-bootstrap look for `libMoltenVK.dylib` on apple platforms, in case the application only has the MoltenVK shared library but not the loader (and the user didn't manually load GetInstanceProcAddr from MoltenVK then give it to vk-bootstrap).

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -50,63 +50,61 @@ bool GenericFeaturesPNextNode::match(GenericFeaturesPNextNode const& requested, 
 class VulkanFunctions {
     private:
     std::mutex init_mutex;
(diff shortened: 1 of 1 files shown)


ADillon1:
This worked on my end @charles-lunarg!

PR #231, opened 2023-11-30
Cleanup CMake finding of Vulkan & installation code 

charles-lunarg:
Adds extensive integration tests that exercise the various ways of finding
the Vulkan-Headers.

vk-bootstrap should now use the Vulkan-Headers or Vulkan::Headers targets if
they were already defined (such as is the case of FetchContent), and will
look for the VulkanHeaders package and Vulkan package as a fallback, and will
FetchContent Vulkan-Headers directly if that fails.

This should make integration seamless with the various ways vulkan-headers
is acquired.

The vk-bootstrap-vulkan-headers target was dropped in favor of directly
linking to Vulkan-Headers (creating a target by that name if none exists).

Update to GLFW 3.3.8 in example code

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,39 +1,57 @@
-cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15)
+
+include(gen/CurrentBuildVulkanVersion.cmake)
+
 project(VulkanBootstrap
     LANGUAGES CXX
-    DESCRIPTION "A Vulkan utility library to ease the initialization steps in Vulkan")
+    DESCRIPTION "A Vulkan utility library to ease the initialization steps in Vulkan"
+    VERSION ${VK_BOOTSTRAP_SOURCE_HEADER_VERSION})
 
 option(VK_BOOTSTRAP_DISABLE_WARNINGS "Disable warnings during compilation" OFF)
 option(VK_BOOTSTRAP_WERROR "Enable warnings as errors during compilation" OFF)
 
-add_library(vk-bootstrap-vulkan-headers INTERFACE)
+if (CMAKE_VERSION VERSION_LESS "3.21")
+    # By default, only build tests & install if this repo is standalone, but still allow configuration by the user
+    string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_TOP_LEVEL)
+endif()
 
-set(VK_BOOTSTRAP_VULKAN_HEADER_DIR "" CACHE STRING "Specify the location of the Vulkan-Headers include directory.")
+option(VK_BOOTSTRAP_TEST "Test Vk-Bootstrap using Catch2 as well as build examples" ${PROJECT_IS_TOP_LEVEL})
+option(VK_BOOTSTRAP_INSTALL "Enable warnings as errors during compilation" ${PROJECT_IS_TOP_LEVEL})
+
+set(VK_BOOTSTRAP_VULKAN_HEADER_DIR "" CACHE FILEPATH "Specify the location of the Vulkan-Headers include directory.")
 mark_as_advanced(VK_BOOTSTRAP_VULKAN_HEADER_DIR)
 
-include(gen/CurrentBuildVulkanVersion.cmake)
+# Check if the user has set this variable explicitly
+if(IS_DIRECTORY ${VK_BOOTSTRAP_VULKAN_HEADER_DIR})
+    add_library(Vulkan-Headers INTERFACE)
+    add_library(Vulkan::Headers ALIAS Vulkan-Headers)
+    target_include_directories(Vulkan-Headers INTERFACE $<BUILD_INTERFACE:${VK_BOOTSTRAP_VULKAN_HEADER_DIR}>)
 
-if(NOT "${VK_BOOTSTRAP_VULKAN_HEADER_DIR}" STREQUAL "")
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -1,39 +1,51 @@
...
+                Vulkan-Headers-for-vk-bootstrap
+                GIT_REPOSITORY https://github.com/KhronosGroup/Vulkan-Headers
+                GIT_TAG        ${VK_BOOTSTRAP_SOURCE_HEADER_VERSION_GIT_TAG}
+            )
+            FetchContent_MakeAvailable(Vulkan-Headers-for-vk-bootstrap)
+        endif()
     endif()
 endif()
+# find_package(Vulkan) only defines Vulkan::Headers, create Vulkan-Headers in this case
+if(TARGET Vulkan::Headers AND NOT TARGET Vulkan-Headers)
```
juan-lunarg, line 41: Why do you want the Vulkan-Headers target instead of Vulkan::Headers?

charles-lunarg: Vulkan::Headers cannot be installed because its an alias, and using `Vulkan-Headers` everywhere simplifies the search logic since I only have to look for 1 target. Like, if Vulkan::Headers was found through `find_package(Vulkan)` the target isn't an alias, but if I use `find_package(VulkanHeaders)`, it *is* an alias (and the install code wasn't happy).

juan-lunarg: > Vulkan::Headers cannot be installed because its an alias

You shouldn't be installing Vulkan::Headers

charles-lunarg: Goes to show what all I know about cmake & installing things! I'll fix it up shortly.

CMakeLists.txt
```diff
@@ -1,39 +1,61 @@
 cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
+
+include(gen/CurrentBuildVulkanVersion.cmake)
+
 project(VulkanBootstrap
     LANGUAGES CXX
-    DESCRIPTION "A Vulkan utility library to ease the initialization steps in Vulkan")
+    DESCRIPTION "A Vulkan utility library to ease the initialization steps in Vulkan"
+    VERSION ${VK_BOOTSTRAP_SOURCE_HEADER_VERSION})
```
juan-lunarg, line 8: I really don't like setting the version like this.

I'd prefer the version be explicit here and make the testing verify the project version is correct.

It's a bit more work but I think it's more elegant solution overall.

CMakeLists.txt
```diff
@@ -1,39 +1,61 @@
 cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
```
juan-lunarg, line 1: FATAL_ERROR isn't needed.

CMakeLists.txt
```diff
@@ -1,39 +1,61 @@
...
 option(VK_BOOTSTRAP_WERROR "Enable warnings as errors during compilation" OFF)
 
-add_library(vk-bootstrap-vulkan-headers INTERFACE)
+if (CMAKE_VERSION VERSION_LESS "3.21")
+    # By default, only build tests & install if this repo is standalone, but still allow configuration by the user
+    if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+        set(PROJECT_IS_TOP_LEVEL ON)
+    else()
+        set(PROJECT_IS_TOP_LEVEL OFF)
+    endif()
```
juan-lunarg, line 19: You can write this logic in 1 line. See Vulkan-Headers

CMakeLists.txt
```diff
@@ -1,39 +1,61 @@
...
+endif()
+
+option(VK_BOOTSTRAP_TEST "Test Vk-Bootstrap using Catch2 as well as build examples" ${PROJECT_IS_TOP_LEVEL})
+option(VK_BOOTSTRAP_INSTALL "Enable warnings as errors during compilation" ${PROJECT_IS_TOP_LEVEL})
 
 set(VK_BOOTSTRAP_VULKAN_HEADER_DIR "" CACHE STRING "Specify the location of the Vulkan-Headers include directory.")
```
juan-lunarg, line 25: Instead of having the type by a STRING it can be a directory. This matters for cmake-gui

CMakeLists.txt
charles-lunarg: How do you set the project version manually? Is it per target or per project?

juan-lunarg:
```
VERSION 1.3.272)
```

Then you test the project version in your test directory.

```
if (NOT PROJECT_VERSION VERSION_EQUAL VK_BOOTSTRAP_SOURCE_HEADER_VERSION)
    message(FATAL_ERROR "Update project version")
endif()
```

charles-lunarg: But how do I update that automatically? Does the update script modify the CMakelists.txt inplace?

charles-lunarg: Looking at `Vulkan-Headers`, I see that it uses `${VK_VERSION_STRING}` in the project().

juan-lunarg (approved):

ext/CMakeLists.txt
```diff
@@ -7,13 +7,13 @@ option(CATCH_BUILD_TESTING "" OFF)
 option(CATCH_ENABLE_WERROR "" OFF)
 option(CATCH_INSTALL_DOCS "" OFF)
 option(CATCH_INSTALL_HELPERS "" OFF)
-option(CATCH_INSTALL_EXTRAS "" ON)
+option(CATCH_INSTALL_EXTRAS "" OFF)
```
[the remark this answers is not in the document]
charles-lunarg: At this point I'm ready to merge, since all the must fixes have been fixed.

PR #249, opened 2024-01-13
Fix: Move objects/vectors where applicable

spnda:
This adds `std::move` or explicit references to some places to avoid unnecessary re-allocations for vectors or other objects. I also replaced a `size` check with a `empty` check and replaced one `push_back` with a `emplace_back`, but there's a lot more of those around the codebase.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1316,7 +1316,7 @@ PhysicalDeviceSelector& PhysicalDeviceSelector::add_desired_extension(const char
     criteria.desired_extensions.push_back(extension);
     return *this;
 }
-PhysicalDeviceSelector& PhysicalDeviceSelector::add_desired_extensions(std::vector<const char*> extensions) {
+PhysicalDeviceSelector& PhysicalDeviceSelector::add_desired_extensions(const std::vector<const char*>& extensions) {
     for (const auto& ext : extensions) {
         criteria.desired_extensions.push_back(ext);
     }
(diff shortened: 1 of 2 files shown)


src/VkBootstrap.cpp
```diff
@@ -1316,7 +1316,7 @@ PhysicalDeviceSelector& PhysicalDeviceSelector::add_desired_extension(const char
     criteria.desired_extensions.push_back(extension);
     return *this;
 }
-PhysicalDeviceSelector& PhysicalDeviceSelector::add_desired_extensions(std::vector<const char*> extensions) {
+PhysicalDeviceSelector& PhysicalDeviceSelector::add_desired_extensions(std::vector<const char*>& extensions) {
```
charles-lunarg, line 1319: Should be `std::vector<const char*> const& extensions` while we are at it.

src/VkBootstrap.h
```diff
@@ -598,7 +598,7 @@ class PhysicalDeviceSelector {
     // Prefer a physical device which supports a specific extension.
     [[deprecated]] PhysicalDeviceSelector& add_desired_extension(const char* extension);
     // Prefer a physical device which supports a set of extensions.
-    [[deprecated]] PhysicalDeviceSelector& add_desired_extensions(std::vector<const char*> extensions);
+    [[deprecated]] PhysicalDeviceSelector& add_desired_extensions(std::vector<const char*>& extensions);
```
charles-lunarg, line 601: use `const&` while we are at it.

charles-lunarg (commented):
Changes are very welcome - initially was hesitant about the std::move() calls but clang-tidy was suggesting the same thing you are, so it must be a good change.

Only one nit, make reference const cause it is. (though, the whole library probably needs to have that applied everywhere consistently...)

PR #251, opened 2024-01-18
Enable Extensions If Present

Diamond-D0gs:
Vulkan has extensions that require other extensions to be enabled to be used, which can lead to the following case.

image

I propose a cleaner solution that can test a vector of extensions and enables all of them if they're all present.

image2


charles-lunarg (approved):
Looks good!
Very good suggestion and appreciate the PR.

PR #253, opened 2024-02-04
Fix: dangling reference in the range-based for loop

SjMxr233:
A dangling reference will apear in the following code:

`
for(auto view : vkb_swapchain.get_image_views().value())
{
      //Anything
}
`

Returning an rvalue reference binding the subobject of a tempory object doesn't make any sense.
Please see [this](https://stackoverflow.com/questions/27368236/return-value-or-rvalue-reference) for more details.
There is a simple example: [https://godbolt.org/z/YcEefWGr7](https://godbolt.org/z/YcEefWGr7)


charles-lunarg (approved):
Thanks for fixing this!

PR #268, opened 2024-03-29
Improve function std::vector arguments, add C++20 std::span support

monomere:
While looking at the code, I've noticed that a lot of functions return a plain `std::vector` not behind a reference, that creates extra copying that could be avoided. I've also moved some code from the cpp file to the header file (like the `CustomQueueDescription` constructor) so that the compiler can inline the code. I think the whole of vk-bootstrap could benefit from that. I should make a separate issue about this.

Also I'm not super sure about the formatting, I did format it with clang-format, but stuff like pointer alignment wasn't changed.

### DIFF
--- a/src/VkBootstrap.h
+++ b/src/VkBootstrap.h
@@ -20,6 +20,16 @@
 #include <cstdio>
 #include <cstring>
 
+#if __cplusplus >= 202002L
+#define VKB_SPAN_OVERLOADS 1
+#elif !defined(VKB_SPAN_OVERLOADS)
+#define VKB_SPAN_OVERLOADS 0
+#endif
+
+#if VKB_SPAN_OVERLOADS
+#include <span>
+#endif
+
 #include <vector>
 #include <string>
 #include <system_error>
@@ -376,8 +386,21 @@ class InstanceBuilder {
     InstanceBuilder& enable_layer(const char* layer_name);
     // Adds an extension to be enabled. Will fail to create an instance if the extension isn't available.
     InstanceBuilder& enable_extension(const char* extension_name);
-    InstanceBuilder& enable_extensions(std::vector<const char*> const& extensions);
+
+    // Add extensions to be enabled. Will fail to create an instance if the extension aren't available.
     InstanceBuilder& enable_extensions(size_t count, const char* const* extensions);
+
+    // Add extensions to be enabled. Will fail to create an instance if the extension aren't available.
+    InstanceBuilder& enable_extensions(std::vector<const char*> const& extensions) {
+        return enable_extensions(extensions.size(), extensions.data());
+    }
+
+#if VKB_SPAN_OVERLOADS
+    // Add extensions to be enabled. Will fail to create an instance if the extension aren't available.
+    InstanceBuilder& enable_extensions(std::span<const char*> extensions) {
+        return enable_extensions(extensions.size(), extensions.data());
+    }
+#endif
 
     // Headless Mode does not load the required extensions for presentation. Defaults to true.
     InstanceBuilder& set_headless(bool headless = true);
@@ -510,7 +533,16 @@ struct PhysicalDevice {
 
     // If all the given extensions are present, make all the extensions be enabled on the device.
     // Returns true if all the extensions are present.
-    bool enable_extensions_if_present(const std::vector<const char*>& extensions);
+    bool enable_extensions_if_present(size_t count, const char* const* extensions);
+    bool enable_extensions_if_present(const std::vector<const char*>& extensions) {
+        return enable_extensions_if_present(extensions.size(), extensions.data());
+    }
+
+#if VKB_SPAN_OVERLOADS
+    bool enable_extensions_if_present(std::span<const char*> extensions) {
+        return enable_extensions_if_present(extensions.size(), extensions.data());
+    }
+#endif
 
     // A conversion function which allows this PhysicalDevice to be used
     // in places where VkPhysicalDevice would have been used.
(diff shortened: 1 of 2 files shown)


src/VkBootstrap.h
```diff
@@ -376,9 +386,22 @@ class InstanceBuilder {
...
+    InstanceBuilder& enable_extensions(std::vector<const char*> const& extensions) {
+        return enable_extensions(extensions.size(), extensions.data());
+    }
+
+#if VKB_SPAN_OVERLOADS
+    // Add extensions to be enabled. Will fail to create an instance if the extension aren't available.
+    InstanceBuilder& enable_extensions(std::span<const char*> extensions) {
+        return enable_extensions(extensions.size(), extensions.data());
+    }
+#endif
```
charles-lunarg, line 403: Why do we have the implementation for the vector & span functions that just forward the arguments to the count + pointer function while the implementation of the count + pointer function is in the .cpp file? 

I would prefer it if all of the implementations were in the .cpp file. While I do think inlining is important, I would much rather make any performance based decision on data, not conjecture.

src/VkBootstrap.h
```diff
@@ -757,7 +811,31 @@ class DeviceBuilder {
...
+        info.queue_descriptions = std::move(queue_descriptions);
+        return *this;
+    }
+
+#if VKB_SPAN_OVERLOADS
+    // For Advanced Users: specify the exact list of VkDeviceQueueCreateInfo's needed for the application.
+    // If a custom queue setup is provided, getting the queues and queue indexes is up to the application.
+    DeviceBuilder& custom_queue_setup(std::span<const CustomQueueDescription> queue_descriptions) {
+        info.queue_descriptions.assign(queue_descriptions.begin(), queue_descriptions.end());
+        return *this;
```
charles-lunarg, line 836: Same as the InstanceBuilder, would much prefer implementations to be in the .cpp

src/VkBootstrap.h
```diff
@@ -802,7 +880,16 @@ struct Swapchain {
...
-    void destroy_image_views(std::vector<VkImageView> const& image_views);
+    void destroy_image_views(size_t count, VkImageView const* image_views);
+    void destroy_image_views(std::vector<VkImageView> const& image_views) {
+        destroy_image_views(image_views.size(), image_views.data());
+    }
+
+#if VKB_SPAN_OVERLOADS
+    void destroy_image_views(std::span<const VkImageView> image_views) {
+        destroy_image_views(image_views.size(), image_views.data());
+    }
```
charles-lunarg, line 891: I'm not seeing the point of this as much, since vk-bootstrap only returns the images in a vector. If tehre was a way to get image views using your own storage then these overloads make more sense.

monomere: You could extract the images into your own storage, and limiting that is a bit pointless imo.

src/VkBootstrap.h
monomere: To answer the first question: so that less code is duplicated

iirc LTO isn't enabled by default on gcc nor clang, which is generally the only way for the compiler to be able to optimize these functions if they weren't inline. Also afaik and it is common practice to put small forwarding functions and simple constructors in the class definition. I could make a small benchmark for this even :)

monomere: Although if you really prefer the function bodies to be in the .cpp file (even if they're a single line), I can change the code

Also I think a great possible performance improvement is the `operator VkXX` methods. It would be the difference between calling a small function (moving stack pointer, etc.) that returns a pointer and using the pointer directly. And since these operators could be used almost everywhere in user code (like passing Device or Instance) it has a measurable impact I'd say.

charles-lunarg: Yes I really would like to see benchmarks - because vkb is an initialization library, very little of this code is called more than once in an application. This library definitely prefers maintainability over performance, and keeping implementations out of the .h file is one way to aid that - as the .h file is in effect the 'documentation'.

src/VkBootstrap.h
charles-lunarg: Very true, so I'll resolve this since its fine as it is (barring moving implementations to .cpp file where appropriate).

src/VkBootstrap.h
monomere: The header files being a sort of documentation is a fair point! (this could be fixed by moving the implementation outside of the class definition, but still in the header file though)

I'll move the method bodies into the .cpp file for now. Although I would still suggest moving the `operator VkXX` methods into the header (it doesn't clutter the code that much, since it fits completely on one line, even with the method declaration)

I'll make the benchmark a bit later, since its not as simple as benchmarking pointer access. But here's a simple [godbolt disassembly](https://godbolt.org/z/PWTEavGxb) (not a whole lot of instructions, so its easy to see the point). Compiled with -O2.
```cpp
extern void do_stuff(int);
void test(A& a) { // A { int unused; int *p; }
    do_stuff(*a.fn1()); // declared out-of-line
    do_stuff(*a.fn2()); // declared inline, returns this->p.
}
```
```asm
test(A&):
  push    rbx
  mov     rbx, rdi
  
  ; this is the out-of-line
  call    A::fn1() ; compiles to "mov rax, qword ptr [rdi+8]; ret"
  mov     edi, DWORD PTR [rax]
  call    do_stuff(int)
  
  ; this is inline
  mov     rax, QWORD PTR [rbx+8]
  mov     edi, DWORD PTR [rax]
  call    do_stuff(int)
```
So even by just analyzing the assembly, you could deduce that in the context of this being used in a context of another function invocation (`vkXX`), the compiler would have to save a lot of registers that store the function arguments and invoke another function.

charles-lunarg (approved):
Wow, I completely forgot about this PR (you might wonder how thats possible, lots of things cause it).

I rebased on main, fixed up the only conflict, and think this is good to go.

src/VkBootstrap.h
```diff
@@ -20,6 +20,16 @@
 #include <cstdio>
 #include <cstring>
 
+#if __cplusplus >= 202002L
```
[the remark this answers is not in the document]
monomere: as far as I understand, no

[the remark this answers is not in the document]
sharadhr: Yes, you'll need `<version>` for C++20 and later, and `<ciso646>` for earlier versions. But the include itself probably needs to be guarded by a C++ version test, something like 

```cpp
#if __cplusplus >= 202002L 
#  include <version>
#endif

// later... 

#if defined(__cpp_lib_span) 
... 
```


`std::span` is a library feature test and those are only defined in the header, unlike *language* feature test macros.

PR #273, opened 2024-04-18
Add enable_features_if_present to PhysicalDevice

### LINKED ISSUE #269: Conditionally enabling device features
Some physical device features can be useful for an application, but not necessary. For example, in my application, I'd like to enable the BC, ASTC, or the ECS texture compression features if they are available. If I would do `add_required_extensions` the physical device selection would fail as often not all three features are available. Therefore, I'd like to see some functionality on the physical device selector or the device builder to specify optional features or features to be enabled, if they are available.

I would propose a function on the device builder that enables a feature if it is available, using `std::mem_fn`. This would look something like this:
```cpp
// The user, enabling a feature
deviceBuilder.enableFeatureIfPresent(std::mem_fn(&VkPhysicalDeviceFeatures::textureCompressionBC));

// Implementation, could also move the std::mem_fn in here.
template <typename Func>
bool vkb::DeviceBuilder::enableFeatureIfPresent(Func function) {
    if (function(availableFeatures)) {
        function(enabledFeatures) = VK_TRUE;
        return true;
    }
    return false;
}
```

@charles-lunarg Does this look like a good idea to you? I would implement a PR for this right now I have your blessing.

Also, another question, but is it intended that `physical_device.features` only represents the enabled features?

charles-lunarg:
Allows users to enable features if they are present, getting back a bool telling them whether the feature is supported and will be enabled on the device.

Fixes #269

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -47,6 +47,54 @@ bool GenericFeaturesPNextNode::match(GenericFeaturesPNextNode const& requested, 
     return true;
 }
 
+void GenericFeaturesPNextNode::combine(GenericFeaturesPNextNode const& right) noexcept {
+    assert(sType == right.sType && "Non-matching sTypes in features nodes!");
+    for (uint32_t i = 0; i < GenericFeaturesPNextNode::field_capacity; i++) {
+        fields[i] = fields[i] || right.fields[i];
+    }
+}
+
+bool GenericFeatureChain::match(GenericFeatureChain const& extension_requested) const noexcept {
+    // Should only be false if extension_supported was unable to be filled out, due to the
+    // physical device not supporting vkGetPhysicalDeviceFeatures2 in any capacity.
+    if (extension_requested.nodes.size() != nodes.size()) {
+        return false;
+    }
+
+    for (size_t i = 0; i < nodes.size() && i < nodes.size(); ++i) {
+        if (!GenericFeaturesPNextNode::match(extension_requested.nodes[i], nodes[i])) return false;
+    }
+    return true;
+}
+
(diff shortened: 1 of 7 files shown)


src/VkBootstrap.cpp
```diff
@@ -863,10 +911,69 @@ std::vector<std::string> check_device_extension_support(
...
+	out.sparseResidencyImage2D = dest.sparseResidencyImage2D || src.sparseResidencyImage2D;
+	out.sparseResidencyImage3D = dest.sparseResidencyImage3D || src.sparseResidencyImage3D;
+	out.sparseResidency2Samples = dest.sparseResidency2Samples || src.sparseResidency2Samples;
+	out.sparseResidency4Samples = dest.sparseResidency4Samples || src.sparseResidency4Samples;
+	out.sparseResidency8Samples = dest.sparseResidency8Samples || src.sparseResidency8Samples;
+	out.sparseResidency16Samples = dest.sparseResidency16Samples || src.sparseResidency16Samples;
+	out.sparseResidencyAliased = dest.sparseResidencyAliased || src.sparseResidencyAliased;
+	out.variableMultisampleRate = dest.variableMultisampleRate || src.variableMultisampleRate;
+	out.inheritedQueries = dest.inheritedQueries || src.inheritedQueries;
+}
```
spnda, line 971: This effectively doesn't do anything, since `out` is modified but that is never returned nor copied anywhere. Did you mean to do `dest = out` at the end?

charles-lunarg: Yep, bugs bugs bugs. This PR definitely isn’t fit in its current state (which is fine, it was a bit rushed)

charles-lunarg: Fixed

charles-lunarg:
I'm going to go ahead and merge - additional features can always be added later.

Issue #277, opened 2024-04-29
Make it possible to check what features have been enabled


n0F4x:
With the help of the recent API `PhysicalDevice::enable_extension_features_if_present` we get a way to enable optional device extension features, but there is no way to ask what features have been enabled afterward.

The same getter is already present for simple device extensions and features.

I presume that adding this new function (`are_extension_features_enabled`) also becomes easier with the addition of `enable_extension_features_if_present`.

charles-lunarg:
Closed as the PR was merged

PR #278, opened 2024-04-30
Add `are_extension_features_present` to `PhysicalDevice`

n0F4x:
Implements #277

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -54,7 +54,7 @@ void GenericFeaturesPNextNode::combine(GenericFeaturesPNextNode const& right) no
     }
 }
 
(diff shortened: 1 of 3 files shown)


charles-lunarg (commented):
Oh, I thought individual feature checking was being asked for - not whether entire feature structs are present. Its much easier to check for structs and not features inside the struct.

I do like what I see here, it implements it cleanly.

charles-lunarg (approved):
I hope you don't mind me going ahead and fixing a bug I found when I added some tests for the new function. Seemed like the best way to get it in, rather than have the normal 'back and forth' of requesting fixes and waiting for them to be implemented. 

The feature was needed, so I appreciate you going ahead and adding it.

Issue #282, opened 2024-05-28
[Fatal Error] [Segmentation Fault] in supports_features Function


imad23m:
Greetings.

I have been experiencing `Segmentation Fault` when starting my project, using GDB I got the following error:

`Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff722992736 in vkb::PhysicalDeviceSelector::is_device_suitable (this=0xb51c3fe5e0, pd=...) at ../dep/vk-bootstrap/src/VkBootstrap.cpp:1189
1189	    bool required_features_supported = detail::supports_features(`

The error happens only sometimes, not every instance, probably related to #91 

I couldn't manage to pin point where the issue in the function call itself happens, however changing all function arguments to pointers seem to have resolved the issue.

it seems sending complex objects in the function call results in some sort of memory issue, but not sure.

if any further information is needed, kindly let me know.

Thank You.

charles-lunarg:
This is a failure that enabling address sanitizer would be the way to debug.

Can you share the code that is causing the error? (most usage of vk-bootstrap is localized to a single file so is easy to share). I'm hesitant to merge a fix for an undiagnosed crash.

charles-lunarg:
Are you willing to try out a different branch of vk-bootstrap?
This uses references instead of passing by value. 
https://github.com/charles-lunarg/vk-bootstrap/tree/supports_features_use_references

PR #283, opened 2024-05-28
[Draft] [Work Around] Fix Segmentation Fault in supports_features Function

imad23m:
This is a draft workaround for the problem mentioned in #282 

The changes include:

-Turn every argument in `support_features` into const pointers.
-Change `support_fueatures` calls to match the new arguments.
-Add checks for `nullptr` in `support_features`.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -969,67 +969,74 @@ void combine_features(VkPhysicalDeviceFeatures& dest, VkPhysicalDeviceFeatures s
 	dest.inheritedQueries = dest.inheritedQueries || src.inheritedQueries;
 }
 
-bool supports_features(VkPhysicalDeviceFeatures supported,
-					   VkPhysicalDeviceFeatures requested,
-					   GenericFeatureChain const& extension_supported,
-					   GenericFeatureChain const& extension_requested) {
-	if (requested.robustBufferAccess && !supported.robustBufferAccess) return false;
-	if (requested.fullDrawIndexUint32 && !supported.fullDrawIndexUint32) return false;
-	if (requested.imageCubeArray && !supported.imageCubeArray) return false;
-	if (requested.independentBlend && !supported.independentBlend) return false;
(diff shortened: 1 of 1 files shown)


charles-lunarg:
Looking at the code, switching to pointers isn't something I would generally opt for. Instead, using references would be my intent, especially since there is no reason for the data to be NULL (especially since the current model of passing by value prevents memory issues). Adding const is a good thing when switching to references (or even pointers for that matter).

imad23m:
Done, anything else?

Notes:
-1) I only did rough testing to see if it works, and it seem it does.
-2) Two of the calls of `supports_features` function in `VkBootstrap.cpp` pass argument of `{}` I am concerned if it will cause any issues. and I haven't done any test regarding these calls.

Thank You.

Imad Laggoune

charles-lunarg (commented):
The {} should be fine because its just empty constructed objects. Its fine if you didn't test this heavily, it isn't a massive change. Do you still see the crash with this change?

charles-lunarg:
Honestly, the 'supports_features' could probably be split into 'supports VkPhysicalDeviceFeatures' and 'supports extension features'. This is not me asking you to do that, just idle thought about code organization (and it would remove the {}'s).

charles-lunarg (approved):
Oops, forgot to click approve, this change does look good after all.

imad23m:
There is one note that may or may not be important:

Checking your `supports_features_use_references` repo commit changes, you added `const&` after VariableType, meanwhile my poll request has `const VariableType &`.

I don't know for sure how significant of a difference they make, I may consider making a change if it turned out one is "better" then the other.

> Do you still see the crash with this change?
No, as far as my testing goes, if crashes started to appear, I will let you know.

>Honestly, the 'supports_features' could probably be split into 'supports VkPhysicalDeviceFeatures' and 'supports extension features'. This is not me asking you to do that, just idle thought about code organization (and it would remove the {}'s).

I agree.

PR #287, opened 2024-06-11
Update README.md

seanbutler:
i think it needs a blank line after heading for a table to render properly


charles-lunarg:
Where is this failing to render? It looks fine for me on github. Now I'm happy to merge this because markdown generally does prefew newlines between stuff, so I'm happy to oblige.

PR #299, opened 2024-08-02
made parameters to set_required_features const

CwTCwT:
this pr restores(?) missing constness of required vulkan feature structs parameters


CwTCwT:
ok i see that s for people who forget to initialize the sType...

PR #300, opened 2024-08-02
PR proposal const& parameters asserting

CwTCwT:
like already mentioned such change will keep/restore(!?) const ness and warn callers not correctly initializing their structs.
sor for multiple commits (you may squash IF you merge) for keeping code style, ...

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1441,20 +1441,20 @@ PhysicalDeviceSelector& PhysicalDeviceSelector::set_required_features(VkPhysical
 }
 #if defined(VKB_VK_API_VERSION_1_2)
 // Just calls add_required_features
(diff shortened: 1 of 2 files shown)


charles-lunarg:
Unfortunately, I will have to reject this PR simply because it changes the API. Before, I didn't require users set the stype, requiring them to do so in a previous commit turned out to break a lot of users unnecessarily. Not to mention they didn't get a nice error, they got cryptic validation messages. 

As much as I'd like to make the parameters const, that isn't really possible with this PR.

charles-lunarg:
I did realize a potential solution that would allow const ref parameter - just create a local copy of the struct and modify that.
Shouldn't be any issue other than a little bit of overhead due to the copy.

charles-lunarg:
Yep, thank you for making the PR otherwise I wouldn't have realized a way to make the functions const (which is very nice!)
https://github.com/charles-lunarg/vk-bootstrap/pull/301

PR #304, opened 2024-08-27
Fix enable_extension_features_if_present on Vulkan 1.0

csmartdalton:
This method should call vkGetPhysicalDeviceFeatures2KHR on Vulkan 1.0.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -1547,10 +1547,18 @@ bool PhysicalDevice::enable_features_node_if_present(detail::GenericFeaturesPNex
     actual_pdf2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
     fill_chain.chain_up(actual_pdf2);
 
-    detail::vulkan_functions().fp_vkGetPhysicalDeviceFeatures2(physical_device, &actual_pdf2);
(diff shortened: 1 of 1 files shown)


src/VkBootstrap.cpp
```diff
@@ -1547,7 +1547,11 @@ bool PhysicalDevice::enable_features_node_if_present(detail::GenericFeaturesPNex
     actual_pdf2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
     fill_chain.chain_up(actual_pdf2);
 
-    detail::vulkan_functions().fp_vkGetPhysicalDeviceFeatures2(physical_device, &actual_pdf2);
+    if (instance_version >= VKB_VK_API_VERSION_1_1) {
+        detail::vulkan_functions().fp_vkGetPhysicalDeviceFeatures2(physical_device, &actual_pdf2);
+    } else {
```
charles-lunarg, line 1552:
```suggestion
    } else if (properties2_ext_enabled) {
```
Good catch! However, to be even more correct the physical device properties 2 extension must be enabled.

csmartdalton: Apologies for the sloppiness! I think this iteration should be complete

charles-lunarg (approved):
Heck ya, did one better than my quick and dirty bool check. 
Thanks for the contribution!

PR #322, opened 2024-11-14
Make double size check more valid

DomNomNom:
It would technically be faster if we just did one of these checks but it seems like there was intent to do some kind of zip(). (referring to python's zip function)

In case someone in the future changes the above if-statement to be a `<` rather than a `!=` then this will be safer.


charles-lunarg (approved):
Yep, thats just a bit of defensive coding which didn't actually get written correctly.

Good eye - I indeed did want to make sure to not iterate past the end of either. Granted, they *should* be the same length before calling this function, but I digress.

PR #328, opened 2025-01-27
Use UINT32_MAX instead of 65536 for QUEUE_INDEX_MAX_VALUE

th3or14:
The sentinel value `QUEUE_INDEX_MAX_VALUE` has been introduced in this https://github.com/charles-lunarg/vk-bootstrap/commit/41203627624b3c21c6cc3e134a73927a17b025f0 commit. Its value 65536 looks confusing to me because the maximum value of `uint32_t` is 4294967295. And I suggest to use an already used in `VkBootstrap.cpp` macro `UINT32_MAX` instead of a magic number 4294967295.

### DIFF
--- a/src/VkBootstrap.h
+++ b/src/VkBootstrap.h
@@ -742,7 +742,7 @@ enum class QueueType { present, graphics, compute, transfer };
 
 namespace detail {
 // Sentinel value, used in implementation only
-inline const uint32_t QUEUE_INDEX_MAX_VALUE = 65536;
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
I think I used 65k because of VK_QUEUE_FAMILY_EXTERNAL, which is ~1U, but now that I look at the value of UINT32_MAX and compare, they are distinct values.

Nothing wrong with using UINT32_MAX since that is much clearer about intent with how common max int is used as a sentinel. No API/ABI issues exist since nothing is specified with anything smaller than 32 bit int.

PR #338, opened 2025-02-08
Check for null device before destroying

spencer-lunarg:
The other destroy functions have this, seem to just be an oversight it is missing for `vkDestroyDevice`


charles-lunarg (approved):
Actually the null checks are redundant. 

> VUID-vkDestroyDevice-device-parameter
> If device is not NULL, device must be a valid [VkDevice](https://registry.khronos.org/vulkan/specs/latest/man/html/VkDevice.html) handle

Its one of those places where its nice to be consistent, so I'll happily merge.

PR #340, opened 2025-02-13
Update example shader names

spencer-lunarg:
for upcoming examples, want to add a new `simple_compute.comp` file and figured we just do a simple 1:1 name mapping


charles-lunarg (approved):
Why didn't I think of that...

Issue #339, opened 2025-02-08
Add quick compute example


spencer-lunarg:
For those who really want to get going fast and don't care about graphics/display I think a `example/simple_compute.cpp` would be cool.

Just something that adds two numbers up, but shows how thin vk-bootstrap can make getting the GPU up and running

PR #342, opened 2025-02-14
Add Simple Compute example

spencer-lunarg:
closes https://github.com/charles-lunarg/vk-bootstrap/issues/339

aka - I can add 2 numbers with now only 300 lines of code


example/simple_compute.cpp
```diff
@@ -0,0 +1,302 @@
...
+    VkDeviceMemory memory_b;
+
+    VkDescriptorSet descriptor_set;
+    VkDescriptorPool descriptor_pool;
+    VkDescriptorSetLayout descriptor_set_layout;
+};
+
+int device_initialization(Init& init) {
+    vkb::InstanceBuilder instance_builder;
+    auto instance_ret = instance_builder.use_default_debug_messenger().request_validation_layers().build();
```
charles-lunarg, line 34: This should have the set_headless() function called on the builder.

example/simple_compute.cpp
```diff
@@ -0,0 +1,302 @@
...
+        std::cout << instance_ret.error().message() << "\n";
+        return -1;
+    }
+    init.instance = instance_ret.value();
+
+    init.inst_disp = init.instance.make_table();
+
+    vkb::PhysicalDeviceSelector phys_device_selector(init.instance);
+    // Skip vk-bootstrap trying to create WSI for you
+    auto phys_device_ret = phys_device_selector.defer_surface_initialization().select();
```
charles-lunarg, line 45: Defer surface initialization isnt for headless, rather its to make vk-bootstrap not do any checking of surface & physical device compatibility. Aka remove it pls thx.

example/simple_compute.cpp
```diff
@@ -0,0 +1,302 @@
...
+int create_descriptor(Init& init, Data& data) {
+    VkDescriptorPoolSize pool_size = { VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 2 };
+
+    VkDescriptorPoolCreateInfo pool_info = {};
+    pool_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_POOL_CREATE_INFO;
+    pool_info.flags = 0;
+    pool_info.maxSets = 1;
+    pool_info.poolSizeCount = 1;
+    pool_info.pPoolSizes = &pool_size;
+    init.disp.createDescriptorPool(&pool_info, nullptr, &data.descriptor_pool);
```
charles-lunarg, line 104: Not even a simple return -1 if not success?

The lack of error checking makes the int return value feel pointless. Could do away with all error checking but like thats a basic rule of thumb - check the return value.

example/simple_compute.cpp
```diff
@@ -0,0 +1,302 @@
...
+    init.disp.getBufferMemoryRequirements(data.buffer_b, &reqs);
+    allocate_info.allocationSize = reqs.size;
+    allocate_info.memoryTypeIndex = get_memory_index(init, reqs.memoryTypeBits);
+    init.disp.allocateMemory(&allocate_info, nullptr, &data.memory_b);
+    init.disp.bindBufferMemory(data.buffer_b, data.memory_b, 0);
+
+    // Update buffers with data
+    // (We are doing the world's most inefficient adding of 42 + 58 to get 100)
+    void* buffer_ptr;
+    init.disp.mapMemory(data.memory_a, 0, VK_WHOLE_SIZE, 0, &buffer_ptr);
```
charles-lunarg, line 219: Kinda of the mind that the memory should be mapped when created and left mapped. No need to free them since its automatic and unnecessary un mapping (only to remap later) is a common perf loss for no gain in capabilities.

charles-lunarg (commented):
Just some api usage & consistency comments.

Otherwise looks great, highlights usage for not graphics.

FYI the queue selection/querying interface has always been a thorn in my side, very happy to hear your thoughts on it.

example/simple_compute.cpp
spencer-lunarg: I guess the example is so simple, this should not fail, not the goal for people to use these bare old ways to create a pool (hope the delete this code for something better)

I guess also wanted to squeeze it in 300 lines lol

charles-lunarg: At that point might as well just remove the error checks anywhere that isnt vkbootstrap. Which honestly is fine, only place worth checking is vkqueuesubmit since everything else should just work.

example/simple_compute.cpp
```diff
@@ -0,0 +1,288 @@
...
+    vkb::destroy_instance(init.instance);
+}
+
+int main() {
+    Init init;
+    Data data;
+
+    if (0 != device_initialization(init)) return -1;
+    if (0 != get_queues(init, data)) return -1;
+    // We are not error checking here, not good practice, but this is just an example program!
```
charles-lunarg, line 280: 😎

charles-lunarg (approved):

Issue #334, opened 2025-01-30
Potential issue of GenericFeaturesPNextNode fields initalization or misunderstanding


th3or14:
I don't understand why fields in `fields[field_capacity]` array of `GenericFeaturesPNextNode` are enabled on initailization by `memset`, could you please explain?

The comment `// Zero out supported features` right above `memset` in `VkBootstrap.cpp` also points to that `memset` was supposed to disable fields. But I assume that setting every byte of `fields` array to `UINT8_MAX` enables all the fields.

th3or14:
This commit https://github.com/charles-lunarg/vk-bootstrap/commit/75db58ae8b4597a6c349f98d8d95a4631b9961cf shows that once the `fields` array initalization code was

```
		for (auto& field : fields) {
			field = 0;
		}
```

And this pull request https://github.com/charles-lunarg/vk-bootstrap/pull/89 introduces `memset`s for the `fields` array initialization. I'm pretty sure that the author intended to zero out the data. I would bring back zero initialization. But I would better replace such `memset`s with explicit assignments to `VK_FALSE` since valid values of `VkBool32` variables are `VK_TRUE` and `VK_FALSE` only.

PR #344, opened 2025-02-15
Fix GenericFeaturesPNextNode fields array initialization

th3or14:
Fixes https://github.com/charles-lunarg/vk-bootstrap/issues/334.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -37,7 +37,13 @@ namespace vkb {
 
 namespace detail {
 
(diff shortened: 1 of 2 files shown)


charles-lunarg (approved):
I honestly cant remember why uint8 max was used instead of zero. I am sure I must of thought it solved a problem but if that isnt documented then thats on me. 

Thank you for the cleanup, and apologies for not responding quicker to your requests. The generic pnext chain logic needs to be redone to use code-gen'd struct sizes, and is just kinda clunky all around.

PR #345, opened 2025-02-16
Use display extension on DIRECT2DISPLAY

kocdemir:
_DIRECT2DISPLAY path probably is not fully supported yet, but this resolves an invalid `VK_KHR_android_surface` usage issue for mock test setup.


charles-lunarg (approved):
Thanks, I don't think I ever actually tested that codepath.

Issue #346, opened 2025-02-16
improve getting started page


spencer-lunarg:
The getting started page is a bit everywhere, we should tidy it up.

First, we should have link to Vulkanized presentation when it comes out.

The starting guide layout I am thinking is:

- Build pattern  
- Return value 
- Instance creation
    - versions
    - extensions
    - layers
- Physical Device Selection
    - Queues
    - extension/features
- WSI 
    - headless
    - simple up and going
- Device creations 
    - make table

Then maybe have an "advanced" page as well with things like

- Instance creation
    - debug callback 
    - layer settings
- Physical Device Selection details
- WSI stuff

@charles-lunarg thoughts?

charles-lunarg:
Is the readme suppose yo be the fist documentation page, or an overview of the project with links to more documentation? Trying to gauge what level of detail you are talking of going into.

charles-lunarg:
Yes, best to introduce the basic stuff first then expand on more complex options and "documentation" like examples later on. Your outline is a fine starting point.

PR #348, opened 2025-02-16
Improve Getting Started doc

spencer-lunarg:
closes https://github.com/charles-lunarg/vk-bootstrap/issues/346

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -1,153 +1,128 @@
 # Getting Started
 
-`vk-bootstrap` reduces the complexity of setting up a vulkan application by simplifying the three initial steps; instance creation, Physical device selection, and device creation.
+`vk-bootstrap` reduces the complexity of dealing with `VkInstance`, `VkPhysicalDevice`, and `VkDevice`, the three things every Vulkan app needs to start.
 
-## Instance Creation
+- [Build Pattern](#build-pattern)
+- [Instance Creation](#instance-creation)
+   * [Versions](#versions)
+   * [Enabling Instance Level Extensions and Layers](#enabling-instance-level-extensions-and-layers)
+- [Physical Device Selection](#physical-device-selection)
+- [Physical Device Object](#physical-device-object)
+- [Device Creation](#device-creation)
+   * [Queues](#queues)
+      + [Custom queue setup](#custom-queue-setup)
+- [Dispatch Table](#dispatch-table)
+- [WSI](#wsi)
+   * [For those who don't want a surface](#for-those-who-dont-want-a-surface)
+   * [Surface Creation](#surface-creation)
+   * [Swapchain](#swapchain)
+- [Cleanup](#cleanup)
+- [Creating your own debug callback](#creating-your-own-debug-callback)
 
-Creating an instance with `vk-bootstrap` uses the `vkb::InstanceBuilder` class.
 
-Simply create a builder variable and call the `build()` member function.
+# Build Pattern
+
+`vk-bootstrap` uses a "build" pattern to creat the objects.
+
+```cpp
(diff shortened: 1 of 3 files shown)


docs/getting_started.md
```diff
@@ -1,154 +1,129 @@
...
 
-// at program end
-vkb::destroy_instance(vkb_instance);
+// Sets VkApplicationInfo::applicationVersion
+instance_builder.set_app_version(3, 0, 0);
+// Sets VkApplicationInfo::engineVersion
+instance_builder.set_engine_version(5, 0, 2);
+
+// The Instance Version (aka the version of your loader)
+// TODO - which?
```
spencer-lunarg, line 77: @charles-lunarg

docs/getting_started.md
```diff
@@ -1,154 +1,129 @@
...
-    std::cerr << "Failed to create Vulkan instance. Error: " << instance_builder_return.error().message() << "\n";
-    return -1;
-}
-```
-Once any possible errors have been dealt with, we can pull the `vkb::Instance` struct out of the `Result`.
-```cpp
-vkb::Instance vkb_instance = instance_builder_return.value();
-```
-This is enough to create a usable `VkInstance` handle but many will want to customize it a bit. To configure instance creation, simply call the member functions on the `vkb::InstanceBuilder` object before `build()` is called.
+// The result also holds the vk-bootstrap wrappre
```
charles-lunarg, line 35: As much as I like the english spelling, that is not right.

docs/getting_started.md
```diff
@@ -1,154 +1,129 @@
...
-```
-Configuration can be chained together and done inline with building, like so.
-```cpp
-auto inst_builder_ret = instance_builder
-        .set_app_name ("Awesome Vulkan Application")
-        .set_engine_name("Excellent Game Engine")
-        .require_api_version(1,0,0)
-        .build();
+// The underlying Vulkan handle can easily be acquired
+vkObject object = wrapper.object;
```
charles-lunarg, line 39:
```suggestion
VkObject object = wrapper.object;
```

docs/getting_started.md
```diff
@@ -191,7 +166,7 @@ Note:
 
 Because `vk-bootstrap` does not manage creating a `VkSurfaceKHR` handle, it is explicitly passed into the `vkb::PhysicalDeviceSelector` for proper querying of surface support details. Unless the `vkb::InstanceBuilder::set_headless()` function was called, the physical device selector will emit `no_surface_provided` error. If an application does intend to present but cannot create a `VkSurfaceKHR` handle before physical device selection, use `defer_surface_initialization()` to disable the `no_surface_provided` error.
```
charles-lunarg, line 167: This line should probably be broken up...

charles-lunarg (commented):
Quick read through while at breakfast. Will try to look harder at it tonight.

So far looks great, save for the TODO. And I want to add stuff about physical device selection but that can be another PR.

docs/getting_started.md
charles-lunarg: Dug through the source code and came up with an answer:

* `require_api_version` says that "1.X must be supported, and set the apiVersion to 1.X"
* `set_minimum_instance_version` says that "1.X must be supported, but don't change the apiVersion.

If you are looking at that and going "wait what, why?". You are thinking what I'm thinking.

I can't see a particular reason to want a high API version unless you specify that you are going to use it. Otherwise, just use the lower version man.

I think I'll make a PR to deprecate it.

spencer-lunarg: (aka, if this is good otherwise, just get this in)

charles-lunarg: Yeah, I'll merge then fixup.

charles-lunarg (approved):
:rocket:

PR #351, opened 2025-02-18
Add SystemInfo::iis_instance_version_available, Better document set_minimum_instance_version

charles-lunarg:
Add SystemInfo::is_instance_version_available
Better document InstanceBuilder::set_minimum_instance_version

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -551,6 +551,15 @@ SystemInfo::SystemInfo() {
             }
         }
     }
+
+    PFN_vkEnumerateInstanceVersion pfn_vkEnumerateInstanceVersion = detail::vulkan_functions().fp_vkEnumerateInstanceVersion;
+
+    if (pfn_vkEnumerateInstanceVersion != nullptr) {
+        VkResult res = pfn_vkEnumerateInstanceVersion(&instance_api_version);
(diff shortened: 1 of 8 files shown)


src/VkBootstrap.cpp
```diff
@@ -819,6 +835,14 @@ InstanceBuilder& InstanceBuilder::set_minimum_instance_version(uint32_t major, u
     info.minimum_instance_version = VKB_MAKE_VK_VERSION(0, major, minor, patch);
     return *this;
 }
+InstanceBuilder& InstanceBuilder::force_api_version(uint32_t major, uint32_t minor) {
```
DethRaid, line 838: I don't have any meaningful suggestions so I'll bikeshed the name

What about `set_api_version`? `force_api_version` makes so... forceful

charles-lunarg: This use case is kinda "advanced" API usage, so a scary name helps ward away unsuspecting users.

spencer-lunarg (approved):
can never have too many ways to set versions in Vulkan!

src/VkBootstrap.cpp
charles-lunarg: I went ahead and removed the function entirely - having a function which did zero checking didn't sit right with me.

PR #352, opened 2025-02-21
Add custom debug callback example

spencer-lunarg:
(want to point in future from Validation Layers update on "how to use future setting in VVL")


charles-lunarg (approved):
Neat and tidy!

Issue #353, opened 2025-02-23
Double destruction of m_value inside Result


th3or14:
A code with a `Result` object like

```
vkb::Result<Sample> result(Sample{});
```

where `Sample` could be like

```
class Sample {
public:
    Sample() { data = new int; }
    ~Sample() { delete data; }
private:
    int* data;
};
```

would lead to calling `delete data` twice. Because if `m_value` inside `Result` is initialized and has a destructor, then such destructor is called twice on a `Result` object destruction due to manual `m_value.~T()` call.

th3or14:
And not only on a `Result` object destruction.

Same double destruction happens inside `Result(Result&& expected)` for a code like

```
vkb::Result<Sample> result(Sample{});
vkb::Result<Sample> anotherResult(std::move(result));
```

To be clear, inside `Result(Result&& expected)` a desctructor of the `expected` object will be called when `expected` goes out of scope due to RAII, and `expected.destroy()` inside `Result(Result&& expected)` adds additional destructor call.

And so on for every other `destroy()` usage I see. I will list other cases with examples.

Inside `Result& operator=(T&& expect)` for a code like

```
    vkb::Result<Sample> result(Sample{});
    Sample sample{};
    result = std::move(sample);
```

Inside `Result& operator=(const T& expect)` for a code like

```
    vkb::Result<Sample> result(Sample{});
    Sample sample{};
    result = sample;
```

Inside `Result& operator=(const Error& error)` for a code like

```
    vkb::Result<Sample> result(Sample{});
    vkb::Error error;
    result = error;
```

Inside `Result& operator=(Error&& error)` for a code like

```
    vkb::Result<Sample> result(Sample{});
    vkb::Error error;
    result = std::move(error);
```

I've updated my pull request. I've also checked all mentioned samples for any memory leakage after removing `destroy()`. All clear, Valgrind reports no leaks.

th3or14:
I think I finally get the logic behind the manual destructor calls, it's due to `union` usage in `Result`. Destructors of `union` members are not expected to be executed automatically. But then I don't understand why double deletion occurs in my examples like if destructors of `union` members are in fact executed automatically.

th3or14:
And memory leaks didn't occur with my fix, though they should occur if destructors of `union` members are not executed automatically. I've closed my pull request and will provide another solution if I get what's going on.

th3or14:
[...]

PR #354, opened 2025-02-23
Eliminate double destruction of m_value

th3or14:
Fixes https://github.com/charles-lunarg/vk-bootstrap/issues/353.

### DIFF
--- a/src/VkBootstrap.h
+++ b/src/VkBootstrap.h
@@ -87,7 +87,6 @@ template <typename T> class Result {
             new (&m_value) T{ std::move(expected.m_value) };
         else
             m_error = std::move(expected.m_error);
-        expected.destroy();
     }
     Result& operator=(Result&& result) noexcept {
         m_init = result.m_init;


src/VkBootstrap.h
```diff
@@ -67,7 +67,7 @@ template <typename T> class Result {
     Result(std::error_code error_code, VkResult result = VK_SUCCESS) noexcept
     : m_error{ error_code, result }, m_init{ false } {}
 
-    ~Result() noexcept { destroy(); }
+    virtual ~Result() noexcept { destroy(); }
```
charles-lunarg, line 70: The `Result` class is not designed nor intended to be a base class. Adding virtual would signal otherwise. Would you mind removing the virtual?

charles-lunarg (commented):
Finally was able to get some time to reproduce the leak locally and can confirm the fix here works and is correct.

While working on this I found that Rule of 5 types don't invoke this behavior, and because vk-bootstrap itself uses a fair number of standard library types (which assuredly follow the rule of 5) I didn't see this for several years.

th3or14:
I tried to implement the Rule of 5 and really stopped getting the double destruction without my fix for a usage like

```
        vkb::Result<Sample> result(Sample{});
        vkb::Result<Sample> anotherResult(std::move(result));
```

after implementing `Sample(Sample&& sample)` this way

```
Sample(Sample&& sample) { data = nullptr; std::swap(sample.data, data); }
```

But it's just because `delete data` is safe to call multiple times if `data` is `nullptr`. One could implement `Sample(Sample&& sample)` in a less optimal yet valid way

```
    Sample(Sample&& sample) { data = new int; std::swap(sample.data, data); }
```

and get the double destruction without my fix.

src/VkBootstrap.h
th3or14: I don't mind, I added it if `Result` is going to be a base class.

charles-lunarg (approved):

PR #358, opened 2025-03-01
Make instance and api versions public in Instance

martty:
This lets users query this before making a device, which might be needed to make decisions on extensions / features.

### DIFF
--- a/src/VkBootstrap.h
+++ b/src/VkBootstrap.h
@@ -318,6 +318,8 @@ struct Instance {
     VkAllocationCallbacks* allocation_callbacks = nullptr;
     PFN_vkGetInstanceProcAddr fp_vkGetInstanceProcAddr = nullptr;
     PFN_vkGetDeviceProcAddr fp_vkGetDeviceProcAddr = nullptr;
+    uint32_t instance_version = VKB_VK_API_VERSION_1_0;
+    uint32_t api_version = VKB_VK_API_VERSION_1_0;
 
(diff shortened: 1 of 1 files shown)


src/VkBootstrap.h
```diff
@@ -318,6 +318,8 @@ struct Instance {
     VkAllocationCallbacks* allocation_callbacks = nullptr;
     PFN_vkGetInstanceProcAddr fp_vkGetInstanceProcAddr = nullptr;
     PFN_vkGetDeviceProcAddr fp_vkGetDeviceProcAddr = nullptr;
+    uint32_t instance_version = VKB_VK_API_VERSION_1_0;
+    uint32_t api_version = VKB_VK_API_VERSION_1_0;
```
charles-lunarg, line 322: Now that the members are public, it'd be good to document *what* they represent, since the naming doesn't obviously indicate what is meant.

```suggestion
    // The apiVersion used to create the instance
    uint32_t instance_version = VKB_VK_API_VERSION_1_0;
    // The instance version queried from vkEnumerateInstanceVersion
    uint32_t api_version = VKB_VK_API_VERSION_1_0;
```

charles-lunarg (commented):
The only hesitation I have is that this would let users "change" the version cause its public. That said, I think most users wouldn't do that. 

I did comment that there should be comments indicating *what* they represent, feel free to revise the suggestions if you wish.

charles-lunarg (approved):

PR #361, opened 2025-03-15
Add VK 1.4 features

DethRaid:
This adds support for specifying Vulkan 1.4 features when selecting a physical device. Tested locally in my hobby renderer


charles-lunarg (approved):
Thanks for fixing this rather obvious omission!

Issue #356, opened 2025-02-25
Memory leaks inside Result


th3or14:
Memory leaks are detected for examples

```
{
    vkb::Result<Sample> result(Sample{});
    vkb::Result<Sample> anotherResult(Sample{});
    result = anotherResult;
}

{
    vkb::Result<Sample> result(Sample{});
    result = vkb::Result<Sample>{Sample{}};
}
```

where `Sample` is

```
class Sample {
public:
    Sample() { data = new int; }
    Sample(Sample const& sample) { data = new int; *data = *sample.data; }
    Sample& operator=(Sample const& sample) { *data = *sample.data; return *this; }
    virtual ~Sample() { delete data; }
private:
    int* data;
};
```

Memory leaks occur because `Result& operator=(Result const& result)` and `Result& operator=(Result&& result)` don't call `destroy()` before assigning new `m_value` or `m_error`. I am ready to suggest a fix after https://github.com/charles-lunarg/vk-bootstrap/pull/354 is accepted.

PR #362, opened 2025-03-19
Eliminate memory leaks inside Result

th3or14:
Fixes https://github.com/charles-lunarg/vk-bootstrap/issues/356.

`destroy()` calls are needed in `Result& operator=(Result const& result)` and `Result& operator=(Result&& result)` because assignment operators, in contrary to constructors, are called for already constructed objects. Constructed objects may have allocated memory that is needed to be freed by calling `destroy()` before proceeding.

You may observe such approach inside other assignment operators: `Result& operator=(const T& expect)`, `Result& operator=(T&& expect)`, `Result& operator=(const Error& error)`, `Result& operator=(Error&& error)`. They all have `destroy()` call in the beginning.

### DIFF
--- a/src/VkBootstrap.h
+++ b/src/VkBootstrap.h
@@ -79,6 +79,7 @@ template <typename T> class Result {
             m_error = expected.m_error;
     }
     Result& operator=(Result const& result) noexcept {
+        destroy();
         m_init = result.m_init;
         if (m_init)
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
I found the same issue when I got address sanitizer working in the repo, but I'll take this so it can be merged right away.

PR #374, opened 2025-04-27
fix triangle example not found shaders

junjie020:
triangle example not foud shader file


charles-lunarg (approved):
Thanks for fixing this!

I seem to always have glslang installed so I never hit this path. I went ahead and modified the shader compile build step so that it just does it in the source tree, matching your change to use shaders in the source tree.

PR #375, opened 2025-04-29
Add support for VK_EXT_layer_settings

thefishlive:
Prevously `vkb::InstanceBuilder` had support for changing settings in VVL via the VK_EXT_validation_features and VK_EXT_validation_flags extensions.

Recently, these two extension have been deprecated and replaced with the VK_EXT_layer_settings extension, which also allows for generic support for settings in any enabled layer.

This change adds support for setting layer settings when creating an instance with `vkb::InstanceBuilder` via the VK_EXT_layer_settings extension.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -660,6 +660,10 @@ Result<Instance> InstanceBuilder::build() const {
         extensions.push_back(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
     }
 
+    if (info.layer_settings.size() > 0) {
+        extensions.push_back(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME);
+    }
+
 #if defined(VK_KHR_portability_enumeration)
     bool portability_enumeration_support =
         detail::check_extension_supported(system.available_extensions, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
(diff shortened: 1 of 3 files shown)


src/VkBootstrap.cpp
```diff
@@ -660,6 +660,10 @@ Result<Instance> InstanceBuilder::build() const {
         extensions.push_back(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
     }
 
+    if (info.layer_settings.size() > 0) {
+        extensions.push_back(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME);
```
thefishlive, line 664: I wasn't sure about enabling the extension on behalf of the user here.

It wasn't previously done for VK_EXT_validation_features / VK_EXT_validation_flags, but since vk-bootstrap was adding the struct to the pnext chain, it seemed like the nicer thing to do for usabilty on the app side

charles-lunarg: Enabling it for the user is the right call. It technically is necessary (even if validation would work without enabling the extension) so better to do the 'obviously correct' thing instead of doing the wrong thing silently.

charles-lunarg (approved):

charles-lunarg:
Thanks for adding support for VK_EXT_layer_settings! Apologies for the delay in getting it in, sometimes I get a little behind.

PR #384, opened 2025-06-19
Fix validation error for using signalled semaphore

### LINKED ISSUE #382: triangle exemple validation error
I run the triangle example and there are validation errors.
Maybe we should avoid having them since the code is made for beginners.

[output.txt](https://github.com/user-attachments/files/20590896/output.txt)

limepixl:
This short PR fixes #382 by allocating enough "finished rendering" semaphores for each swapchain image in the triangle example. This silences the validation error `VUID-vkQueueSubmit-pSignalSemaphores-00067`.


charles-lunarg (approved):
Would have fixed it the exact same way! Thanks for talking this issue.

PR #385, opened 2025-06-21
Small change in combine_features and supports_features, reduces amount of hand-written code.

ZloyKorovanovich:
VkPhysicalDeviceFeatures struct can be represented just as an array of VkBool32, so instead of checking for each single member, we can run a for loop with fixed number of iterations.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -957,124 +957,22 @@ std::vector<std::string> check_device_extension_support(
 
 // clang-format off
 void combine_features(VkPhysicalDeviceFeatures& dest, VkPhysicalDeviceFeatures src){
-    dest.robustBufferAccess = dest.robustBufferAccess || src.robustBufferAccess;
-	dest.fullDrawIndexUint32 = dest.fullDrawIndexUint32 || src.fullDrawIndexUint32;
-	dest.imageCubeArray = dest.imageCubeArray || src.imageCubeArray;
-	dest.independentBlend = dest.independentBlend || src.independentBlend;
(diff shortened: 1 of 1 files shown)


th3or14:
I think that accessing an object of effective type `VkPhysicalDeviceFeatures` through a pointer to `VkBool32` is undefined behavior because it violates strict aliasing rules. https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8 A standard-compliant way to solve it could be creating extra `VkBool32` arrays, then `memcpy`ing data from `VkPhysicalDeviceFeatures` objects to those arrays, and only then comparing array elements in a loop. But this way would involve extra copies.

Additionally, I don't like relying on that fields of `VkPhysicalDeviceFeatures` are always `VkBool32` and always packed, even though it may always be true. I rather don't like relying on beliefs of how data is stored in memory.

ZloyKorovanovich:
> I think that accessing an object of effective type `VkPhysicalDeviceFeatures` through a pointer to `VkBool32` is undefined behavior because it violates strict aliasing rules. https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8 A standard-compliant way to solve it could be creating extra `VkBool32` arrays, then `memcpy`ing data from `VkPhysicalDeviceFeatures` objects to those arrays, and only then access array elements in a loop. But this way would involve extra copies.
> 
> Additionally, I don't like relying on that fields of `VkPhysicalDeviceFeatures` are always `VkBool32` and always packed, even though it may always be true. But I rather don't like relying on beliefs of how data is stored in memory.

Oh I see, thanks for reference, I read it. I understand your point now, yeah that might be the case.

PR #387, opened 2025-06-23
Replace homebrew codegen with VulkanObject

charles-lunarg:
Replace the custom parsing of xml using xmltodict with the purpose build vulkan_object.py package. This allows dumping all of the xml parsing logic, as that is now handled in vulkan_object for us.

Because VulkanObject is a different code generator, it doesn't result in the same order in the generated output. To develop this commit, the prior codegen sorted the commands by name then staged the changes, then the new codegen also sorted the commands and was run. This creates a useable diff that shows minor changes.

The notable change is only using the command.extension list as the 'dependencies' list. The logic for the dependencies was somewhat faulty, as the header files do not follow the dependencies as far as defining the types is concerned. The depends list is for  runtime, not compile time.


.github/workflows/run_autogen.yml
```diff
@@ -39,18 +39,17 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-            python-version: '3.9'
+            python-version: '3.10'
 
-      - name: Install python package 'xmltodict'
-        run: pip install xmltodict
+      - run: pip install vulkan_object
```
spencer-lunarg, line 44: :100: for doing this the "python way"

spencer-lunarg (approved):
can't see the generated output (assuming its white space noise?)

the python looks WAY simpler

Issue #363, opened 2025-03-22
Strict aliasing rules violation


th3or14:
I'm compiling with GCC.

When compiling with `-Wstrict-aliasing=1` warnings level enabled and at least `-O2` optimizations level, I see multiple warnings like `warning: dereferencing type-punned pointer might break strict-aliasing rules`. For example, for the line

```
pNext_chain.push_back(reinterpret_cast<VkBaseOutStructure*>(&messengerCreateInfo));
```

in `VkBootstrap.cpp`. I looked closely and confirmed that accessing data through `VkBaseOutStructure` pointers may be dangerous when data has a type other than `VkBaseOutStructure`.

I wrote a reproducer for the undefined behavior. Something like the first 3 lines of code is happening in `VkBootstrap.cpp`, next I'm attempting to abuse the strict aliasing violation.

```
std::vector<VkBaseOutStructure*> chain;
VkDebugUtilsMessengerCreateInfoEXT info = {};
chain.push_back(reinterpret_cast<VkBaseOutStructure*>(&info));

// set pNext to an arbitrary value 1
info.pNext = reinterpret_cast<const void *>(1);
// nullify the same pNext using a pointer to VkBaseOutStructure
chain[0]->pNext = nullptr;

printf("Value of pNext obtained through info: %p.\n", info.pNext);
printf("Value of pNext obtained through chain: %p.\n", chain[0]->pNext);
bool areAddressesEqual = reinterpret_cast<const void *>(&info.pNext) == reinterpret_cast<const void *>(&chain[0]->pNext);
printf("Are pNext addresses physically equal in memory? %s.\n", (areAddressesEqual ? "Yes" : "No"));
```

Compiling and running with at least `-O2` optimizations level gives the output

```
Value of pNext obtained through info: 0x1.
Value of pNext obtained through chain: (nil).
Are pNext addresses physically equal in memory? Yes.
```

because accessing data through `chain[0]->pNext` is undefined behavior here. Even though `pNext` addresses are physically equal in memory.

Adding `-fno-strict-aliasing` to the compilation flags changes the output to

```
Value of pNext obtained through info: (nil).
Value of pNext obtained through chain: (nil).
Are pNext addresses physically equal in memory? Yes.
```

because now the compiler knows that the strict aliasing rules may be violated and doesn't apply related optimizations.

[...]

PR #393, opened 2025-07-09
Rewrite features struct pNext chain implementation

charles-lunarg:
The previous implementation had multiple instances of UB, as it contained unsound type punning, and issues with padding bytes causing differing behavior.

This new implementation now has generated code to handle comparing and merging the fields within the struct, guaranteeing no issues with padding. Structs in a chain are all stored in a std::vector<std::byte> array. The location of structs are stored in a separate struct, and when accessed use reinterpret_cast to convert a pointer to the start of the struct to the correct type. When only accessing the sType or pNext, memcpy is used to convert the type into a VkBaseOutStructure first, and if writing is required, memcpy'd back onto the storage vector.


n0F4x (commented):
I read through the most important changes in VkBootstrap.h and VkBootstrap.cpp.
This PR feels unfinished and still error-prone because of all the `pNext` pointer invalidation (but that may just be how life intended them to be in this case).

th3or14:
Thank you for pinging me. I'm unable to build on branch `feature_chain` at the moment due to a lot of errors like

```
In file included from VkBootstrap.h:30,
                 from VkBootstrap.cpp:17:
VkBootstrapFeatureChain.h:974:102: error: ‘VkPhysicalDeviceShaderBfloat16FeaturesKHR’ has not been declared
  974 | ine void compare_VkPhysicalDeviceShaderBfloat16FeaturesKHR(std::vector<std::string> & error_list, VkPhysicalDeviceShaderBfloat16FeaturesKHR const& supported, VkPhysicalDeviceShaderBfloat16FeaturesKHR const& requested) {
      |                                                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

for various types. I suppose that it's due to lack of guards like `#if defined(VK_KHR_shader_bfloat16)`.

charles-lunarg (commented):
Very helpful review, identified that there was a certain amount of muck still left in the implementation after the first draft(s) using code gen.

Major changes are to remove unnecessary functions in FeaturesChain and splitting the implementation & declarations of the generated code to reduce cost of including the headers.

charles-lunarg:
> Thank you for pinging me. I'm unable to build on branch `feature_chain` at the moment due to a lot of errors like
> 
> ```
> In file included from VkBootstrap.h:30,
>                  from VkBootstrap.cpp:17:
> VkBootstrapFeatureChain.h:974:102: error: ‘VkPhysicalDeviceShaderBfloat16FeaturesKHR’ has not been declared
>   974 | ine void compare_VkPhysicalDeviceShaderBfloat16FeaturesKHR(std::vector<std::string> & error_list, VkPhysicalDeviceShaderBfloat16FeaturesKHR const& supported, VkPhysicalDeviceShaderBfloat16FeaturesKHR const& requested) {
>       |                                                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> for various types. I suppose that it's due to lack of guards like `#if defined(VK_KHR_shader_bfloat16)`.

I added a compile time option `VK_BOOTSTRAP_FORCE_FETCH_CONTENT` that forces VkBootstrap to acquire Vulkan-Headers using FetchContent (using the version the codegen used). This fixes that compile issue for me, so should help you be able to review it.

CMakeLists.txt
```diff
@@ -67,7 +72,7 @@ endif()
 
 if(NOT VK_BOOTSTRAP_DISABLE_WARNINGS)
     if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|AppleClang" OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND VK_BOOTSTRAP_COMPILER_FRONTEND MATCHES "GNU"))
-        target_compile_options(vk-bootstrap-compiler-warnings INTERFACE -Wall -Wextra -Wconversion -Wsign-conversion)
+        target_compile_options(vk-bootstrap-compiler-warnings INTERFACE -Wall -Wextra -Wconversion -Wsign-conversion -Wstrict-aliasing=1)
```
th3or14, line 75: It seems that `-Wstrict-aliasing=1` makes no sense without at least `-O2`. Or, to be precise, without `-fstrict-aliasing`, which is enabled by `-O2` or above by default.

So if I add `-fstrict-aliasing` or at least `-O2`, then I see a lot of `warning: dereferencing type-punned pointer might break strict-aliasing rules [-Wstrict-aliasing]`.

charles-lunarg: Sounds like i have more changes to make!

th3or14: If it takes too much efforts, there is still an option to just keep related optimizations explicitly disabled by adding `-fno-strict-aliasing`. Vulkan-Samples did so. https://github.com/KhronosGroup/Vulkan-Samples/issues/1207

If users build with your set of compile options, which even right now do not enable related optimizations, users are probably safe. Adding `-fno-strict-aliasing` will make sure that if users add `-O2` or above, they are still safe.

The only drawback I see is that if users build with a different set of compile options, that might be not safe. For example, if users don't use your `CMakeLists.txt` and don't pay attention to your compile options, following the copy-paste setting up method (https://github.com/charles-lunarg/vk-bootstrap?tab=readme-ov-file#copy-paste).

charles-lunarg: Oh an my stance is that Vk-bootstrap should be clean as a whistle, so that no matter what flags users turn on, they do not get anything from it. Sure that may not happen in practice (I am not militant about it), but the policy means that I should enable all the warnings I can, fix them, and do not require users to set specific compiler options in order to use the library. I want to limit the complexity of using vk-bootstrap outside of CMake, so minimal flags and library depedencies is essential.

charles-lunarg:
I'm eager to get this in, as the PR fixes things that are fundamentally broken. Sure they may work for many users, but not all. And while I'm sure there is *some* bug in this PR, I don't think it'll be found without first merging and getting people using it.

charles-lunarg:
@th3or14 @n0F4x Feel free to comment here if you do find issues, or create a new issue to track it, as I'm going ahead and merging it. I can always undo the commit if there is a game breaking bug found.

CMakeLists.txt
th3or14: I used `-Wstrict-aliasign=1` because higher levels didn't produce warnings in, for example, this https://github.com/charles-lunarg/vk-bootstrap/issues/363#issue-2940172787 reproducer, for which warnings were not false positives.

If I compile with `-Wstrict-aliasign=1` now, then there are no warnings for the files under `vk-bootstrap/src`, that's great. There are a lot of warnings originating to `vk-bootstrap/build/_deps/vulkan-headers-for-vk-bootstrap-src/include/vulkan/vulkan.hpp`. I ignore them because they are from dependencies. And there are 2 warnings at `vk-bootstrap/example/simple_compute.cpp` that look false positive to me in this case. I ran `vk-bootstrap-simple_compute` with enabled strict aliasing optimizations and got the expected result. I also didn't manage to abuse the strict aliasing violation in `vk-bootstrap-simple_compute`, if takes place.

Issue #394, opened 2025-07-10
Build error due to a probably wrong type of pReleaseInfo


th3or14:
I'm struggling to build on branch `main`.

```
In file included from VkBootstrap.h:29,
                 from VkBootstrap.cpp:17:
VkBootstrapDispatch.h:5835:46: error: ‘VkReleaseSwapchainImagesInfoKHR’ does not name a type; did you mean ‘VkReleaseSwapchainImagesInfoEXT’?
 5835 |     VkResult releaseSwapchainImagesEXT(const VkReleaseSwapchainImagesInfoKHR* pReleaseInfo) const noexcept {
      |                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                              VkReleaseSwapchainImagesInfoEXT
```

I discovered that I have `VK_EXT_swapchain_maintenance1` defined while I don't have `VK_KHR_swapchain_maintenance1` defined.

I also see that in my Vulkan headers I have `VkReleaseSwapchainImagesInfoEXT` structure defined while I don't have `VkReleaseSwapchainImagesInfoKHR` structure defined.

So I believe that the function

```
#if (defined(VK_EXT_swapchain_maintenance1))
    VkResult releaseSwapchainImagesEXT(const VkReleaseSwapchainImagesInfoEXT* pReleaseInfo) const noexcept {
        return fp_vkReleaseSwapchainImagesEXT(device, pReleaseInfo);
    }
#endif
```

should use `VkReleaseSwapchainImagesInfoEXT*` instead of `VkReleaseSwapchainImagesInfoKHR*` for the type of `pReleaseInfo`.

Surprisingly, the documentation for the function `vkReleaseSwapchainImagesEXT` disagrees. https://registry.khronos.org/vulkan/specs/latest/man/html/vkReleaseSwapchainImagesEXT.html But it could be a mistake in the documentation.

charles-lunarg:
Gah, I had hoped this was just a problem due to using older headers with a vk-bootstrap that expects something newer, but alas it isn't so. I have been swamped with other tasks so haven't been able to look into this fully. (and completely forgot about it to be quite clear, lots of emails means some get lost).

One solution is to hardcode it in the codegen to use VkReleaseSwapchainImagesInfoEXT instead. I don't like that long term, but would be sufficient for unblocking users.

[...]

PR #395, opened 2025-07-10
Correct the type of pReleaseInfo in releaseSwapchainImagesEXT

th3or14:
Fixes https://github.com/charles-lunarg/vk-bootstrap/issues/394.

Editing `src/VkBootstrapDispatch.h` also fixed "No newline at end of file".


th3or14:
I see now that `src/VkBootstrapDispatch.h` is a generated file. It's not good to manually modify it as I did at first.

Alright, I attempted to modify the code generation in an expressive way. The newly updated `src/VkBootstrapDispatch.h` is generated by the modified script.

th3or14:
But maybe it should be fixed in `VulkanObject` or `vk.xml` rather than in the code generation here.

PR #406, opened 2025-09-07
Replace union with std::variant in Result

th3or14:
No more manual and conditional destructor calls. No more `m_init` flag based conditional initializations. No more being unhappy with default destructor, default copy constructors, default assignment operators.

API is preserved.


th3or14:
Additionally, I removed `noexcept` for a few methods because `std::get` throws `std::bad_variant_access` on errors. I also removed redundant `assert`s because the exceptions are thrown anyway there.

charles-lunarg (approved):
C++17 to the rescue!

I wrote this result class because I was targeting C++11 when I started this library in 2020, which is I believe a reasonable baseline. But that was 5 years ago, and C++17 is widespread. So great idea to remove the jank of a custom union type without changing a thing.

I went ahead and rebased on main to hopefully get it passing tests.

Issue #407, opened 2025-09-07
CI builds use gcc when clang is implied


th3or14:
While I was troubleshooting CI builds, I noticed that `Linux (Debug, clang, clang++)` and `Linux (Release, clang, clang++)` use gcc instead of clang. For example, if you expand `CMake Configure` step of this https://github.com/charles-lunarg/vk-bootstrap/actions/runs/17525289412/job/49774855894?pr=406 `Linux (Debug, clang, clang++)` build, you will see the line

```
 -- The CXX compiler identification is GNU 13.3.0
```

For a reference, when I locally configure with clang, I see the line

```
-- The CXX compiler identification is Clang 18.1.8
```

PR #408, opened 2025-09-08
Specify compilers in CI

th3or14:
Fixes https://github.com/charles-lunarg/vk-bootstrap/issues/407.


charles-lunarg (approved):
Whoops!
I bet I did the classic error of seeing CI say something and not actually double checking that it *does* the thing it says it's doing.

Issue #377, opened 2025-05-02
Mutex lock error when building Vulkan Instance


Siddharth2077:
Here's my simple code for creating a Vulkan Instance (I'm following the vkguide.dev tutorials)
```
void VulkanEngine::init_vulkan() {
    vkb::InstanceBuilder vulkanInstanceBuilder;

    // Create a Vulkan instance with basic debugging using the builder
    auto instanceBuilderReturn = vulkanInstanceBuilder
        .set_app_name("Vulkan Game Engine")
        .require_api_version(1, 3, 0)
#ifndef NDEBUG
        .request_validation_layers()
#endif
        .use_default_debug_messenger()
        .build();

    if (!instanceBuilderReturn) {
        std::cout << "ERROR in creating Vulkan Instance\n";
    }

    // Created the Vulkan instance successfully
    vkb::Instance vkbInstance = instanceBuilderReturn.value();
    m_vulkanInstance = vkbInstance.instance;
    m_vulkanDebugMessenger = vkbInstance.debug_messenger;
}
```

When I run my project in DEBUG mode, it fails in the ``` build() ``` method of the instance builder. (In Release mode, I do not get this error).

Tried adding break points and it fails in the very first line of the ``` build() ``` method:

```
Result<Instance> InstanceBuilder::build() const {

    auto sys_info_ret = SystemInfo::get_system_info(info.fp_vkGetInstanceProcAddr);
```

The following is the code within ``` get_system_info ``` method:
```
Result<SystemInfo> SystemInfo::get_system_info(PFN_vkGetInstanceProcAddr fp_vkGetInstanceProcAddr) {
    // Using externally provided function pointers, assume the loader is available
    if (!detail::vulkan_functions().init_vulkan_funcs(fp_vkGetInstanceProcAddr)) {
        return make_error_code(InstanceError::vulkan_unavailable);
    }
    return SystemInfo();
}
```

It then enters the ``` init_vulkan_funcs ``` function which is the cause of the Failure...

```
bool init_vulkan_funcs(PFN_vkGetInstanceProcAddr fp_vkGetInstanceProcAddr = nullptr) {
    std::lock_guard<std::mutex> lg(init_mutex);
    if (fp_vkGetInstanceProcAddr != nullptr) {
        ptr_vkGetInstanceProcAddr = fp_vkGetInstanceProcAddr;
    } else {
        bool ret = load_vulkan_library();
        if (!ret) return false;
    }
```

[...]

PR #414, opened 2025-09-13
Working on tests

th3or14:
1. I removed the irrelevant code

```
    vkb::InstanceBuilder builder;
    auto ret = builder.build();
    REQUIRE(ret);
```

from 2 tests. For example, gettng the system info in the following test

```
TEST_CASE("SystemInfo Loading Vulkan Automatically", "[VkBootstrap.loading]") {
    [[maybe_unused]] VulkanMock& mock = get_and_setup_default();
    auto info_ret = vkb::SystemInfo::get_system_info();
    REQUIRE(info_ret);
    vkb::InstanceBuilder builder;
    auto ret = builder.build();
    REQUIRE(ret);
}
```

prior to instance building doesn't affect instance building. The instance builder gets the system info internally from scratch anyway:

```
Result<Instance> InstanceBuilder::build() const {

    auto sys_info_ret = SystemInfo::get_system_info(info.fp_vkGetInstanceProcAddr);
```

I supposed that maybe the intention of additional instance building was to obtain `detail::vulkan_functions().ptr_vkGetInstanceProcAddr`, which is set after calling `vkb::SystemInfo::get_system_info()`, and use it for instance building like

```
vkb::InstanceBuilder builder{ detail::vulkan_functions().ptr_vkGetInstanceProcAddr };
```

but the whole `detail::vulkan_functions()` is not exposed to users, hidden inside `src/VkBootstrap.cpp`. So I removed irrelevant instance building happening after testing `SystemInfo`.

2. In an attempt to do something with https://github.com/charles-lunarg/vk-bootstrap/issues/377, I added a test for running in multiple threads the code that is expected to be thread safe. It uses the mentioned line `std::lock_guard<std::mutex> lg(init_mutex);` inside.

3. I attempted to unify the usage of `std::vector<char>`.

### DIFF
--- a/tests/vulkan_mock.hpp
+++ b/tests/vulkan_mock.hpp
@@ -2,6 +2,7 @@
 
 #include <assert.h>
 #include <cstdint>
+#include <cstring>
 
 #include <memory>
 #include <utility>
@@ -10,6 +11,27 @@
 #include <vector>
 
 #include <vulkan/vulkan_core.h>
+
+typedef std::vector<char> SerializedStruct;
+
+inline SerializedStruct create_serialized_struct_from_pointer(const void *input_data, size_t input_size) {
+    if (static_cast<const VkBaseOutStructure*>(input_data)->sType == 0) {
+        throw std::runtime_error(
+            "create_serialized_struct_from_pointer being passed in a struct without setting the sType!");
+    }
+    SerializedStruct new_struct;
+    new_struct.resize(input_size);
+    std::memcpy(new_struct.data(), input_data, new_struct.size());
+    return new_struct;
+}
+
+template <typename T> SerializedStruct create_serialized_struct_from_object(T object) {
+    if (object.sType == 0) {
+        throw std::runtime_error(
+            "create_serialized_struct_from_object being passed in a struct without setting the sType!");
+    }
+    return create_serialized_struct_from_pointer(&object, sizeof(T));
+}
 
 // Helper function to return the size of the sType if it is a known features struct, otherwise return 0
 // Hand written, must be updated to include any used struct.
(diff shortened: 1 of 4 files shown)


tests/vulkan_mock.hpp
```diff
@@ -11,6 +12,26 @@
 
 #include <vulkan/vulkan_core.h>
 
+using SerializedStruct = std::vector<char>;
+
+inline SerializedStruct create_serialized_struct_from_pointer(const void* input_data, size_t input_size) {
+    if (static_cast<const VkBaseOutStructure*>(input_data)->sType == 0) {
```
charles-lunarg, line 18: I want to say this is UB, since input_data might not be a pointer to `VkBaseOutStructure`, but at the same time because this is test code and I'm fairly certain this will just work I don't want to bother you to fix it.

charles-lunarg (approved):
Good cleanup. 

The test code used to use the vk-bootstrap internal code, but at some point I realized that it didn't offer any real help since the test usage is simple compared to what vk-bootstrap itself needs. That might have been at the time I got the mock going, and found I needed to rewrite it for the mock anyhow. Kudos for cleaning it up with a good typedef and helper functions.

tests/vulkan_mock.hpp
th3or14: I wonder if changing `static_cast` to `dynamic_cast` would be sufficient.

th3or14: `dynamic_cast` doesn't suit. I'm working on something like

```
        VkBaseOutStructure out_structure{};
        memcpy(&out_structure, structs.at(i), sizeof(VkBaseOutStructure));
#if !defined(NDEBUG)
        assert(out_structure.sType != VK_STRUCTURE_TYPE_APPLICATION_INFO);
#endif
```

charles-lunarg (approved):
Perfect way to avoid needing a bunch of tedious memcpy's.
Test code has to modify the mock to add new features structs anyhow, so fine with me if there is a hard coded stype thing.

PR #415, opened 2025-09-13
Handle when minimum_instance_version is set but required_api_version isn't

charles-lunarg:
Adds missing logic to handle when set_minimum_instance_version is called but require_api_version isn't. 

Now correctly logs error messages when set_minimum_instance_version results in an instance version error.

Adds tests to check the validity of set_minimum_instance_version.

### DIFF
--- a/tests/bootstrap_tests.cpp
+++ b/tests/bootstrap_tests.cpp
@@ -205,6 +205,55 @@ TEST_CASE("Device Configuration", "[VkBootstrap.bootstrap]") {
     vkb::destroy_instance(instance);
 }
 
+TEST_CASE("Physical Device Version", "[VkBootstrap.minimum_device_version]") {
+    VulkanMock& mock = get_and_setup_default();
+    mock.instance_api_version = VK_API_VERSION_1_4;
+    mock.physical_devices_details[0].properties.apiVersion = VK_API_VERSION_1_1;
+    mock.physical_devices_details[0].properties.deviceID = 1;
+    add_basic_physical_device(mock);
+    mock.physical_devices_details[1].properties.apiVersion = VK_API_VERSION_1_4;
+    mock.physical_devices_details[1].properties.deviceID = 4;
+
+    auto instance_ret = vkb::InstanceBuilder{}.set_headless().require_api_version(1, 4).build();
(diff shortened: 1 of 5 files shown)


tests/bootstrap_tests.cpp
```diff
@@ -555,14 +607,71 @@ TEST_CASE("Minimum instance API version", "[VkBootstrap.api_version]") {
...
+    REQUIRE(ret_1_4.error() == vkb::InstanceError::vulkan_version_1_4_unavailable);
+}
+TEST_CASE("Minimum instance API version without required api version", "[VkBootstrap.set_minimum_instance_version]") {
+    VulkanMock& mock = get_and_setup_default();
+    mock.instance_api_version = VK_API_VERSION_1_2;
+
+    auto ret = vkb::InstanceBuilder{}.set_headless().set_minimum_instance_version(1, 2).build();
+    REQUIRE(ret);
+    REQUIRE(mock.api_version_set_by_vkCreateInstance == VK_API_VERSION_1_2);
+    mock.api_version_set_by_vkCreateInstance = 0;
```
th3or14, line 643: I see no point in manual resetting `api_version_set_by_vkCreateInstance` because `get_and_setup_default()` will recreate the internal `mock` anyway. Manual resetting here makes think that it makes sense and that it doesn't happen otherwise, like in case of `vk_lib.close()`.

tests/bootstrap_tests.cpp
```diff
@@ -555,14 +607,71 @@ TEST_CASE("Minimum instance API version", "[VkBootstrap.api_version]") {
                        .set_minimum_instance_version(VK_MAKE_API_VERSION(0, 1, 2, 0))
                        .build();
         REQUIRE(ret);
+        REQUIRE(mock.api_version_set_by_vkCreateInstance == VK_API_VERSION_1_4);
+        mock.api_version_set_by_vkCreateInstance = 0;
```
th3or14, line 611: Same here, there are no further `mock` usages after leaving this scope, there is no need to manually reset `api_version_set_by_vkCreateInstance`. What do you think about one manual resetting between 2 scopes?

th3or14 (approved):

PR #417, opened 2025-09-14
Fix threading problems

charles-lunarg:


th3or14 (approved):
I see no issues with the pull request. I only needed to disable randomization of the virtual address space on my system in order to avoid

```
FATAL: ThreadSanitizer: unexpected memory mapping
```

There is no clang thread sanitizer pipeline added, so I tested the thread sanitizer with clang as well. I can also reproduce the threading problems after rolling back the fix.

charles-lunarg:
> clang thread sanitizer pipeline added

Oh, hmm I figured just GCC would be enough as I thought it was the same codebase. But maybe not, and should be added.

PR #418, opened 2025-09-16
Fix threading bug in test

charles-lunarg:


th3or14:
Another possible option is to make `api_version_set_by_vkCreateInstance` atomic variable.

charles-lunarg:
> Another possible option is to make api_version_set_by_vkCreateInstance atomic variable.

I thought about that briefly. If I am being honest, it would make more sense to create a 'per-thread' mock object so that each thread can do the things it needs. But thats not a pressing need so simple 'just don't write to api_version_set_by_vkCreateInstance when not needed' should suffice.

PR #420, opened 2025-09-19
Cleanup image views in error path

charles-lunarg:
Add a simple test to ensure that all images are cleaned up. Not a perfect test, but is easy enough to verify the function was called enough times.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -2038,8 +2038,11 @@ Result<std::vector<VkImageView>> Swapchain::get_image_views(const void* pNext) {
         createInfo.subresourceRange.baseArrayLayer = 0;
         createInfo.subresourceRange.layerCount = 1;
         VkResult res = internal_table.fp_vkCreateImageView(device, &createInfo, allocation_callbacks, &views[i]);
-        if (res != VK_SUCCESS)
+        if (res != VK_SUCCESS) {
+            // Cleanup already created image views
+            destroy_image_views(i, views.data());
             return Result<std::vector<VkImageView>>{ SwapchainError::failed_create_swapchain_image_views, res };
+        }
     }
     return views;
 }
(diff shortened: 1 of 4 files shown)


tests/vulkan_mock.hpp
```diff
@@ -105,6 +105,8 @@ struct VulkanMock {
         physical_devices_details.emplace_back(std::move(details));
     }
 
+    uint32_t created_image_view_count = 0;
+    uint32_t fail_image_creation_on_iteration = 0; // 0 means do not fail instance creation
```
th3or14, line 109: If 0 means do not fail instance creation, then you can't fail on the first (0) iteration. I would use `UINT32_MAX` instead of 0 as an "off" value.

src/VkBootstrap.cpp
```diff
@@ -2038,8 +2038,13 @@ Result<std::vector<VkImageView>> Swapchain::get_image_views(const void* pNext) {
         createInfo.subresourceRange.baseArrayLayer = 0;
         createInfo.subresourceRange.layerCount = 1;
         VkResult res = internal_table.fp_vkCreateImageView(device, &createInfo, allocation_callbacks, &views[i]);
-        if (res != VK_SUCCESS)
+        if (res != VK_SUCCESS) {
+            // Cleanup already created image views
+            for (size_t j = 0; j < i; j++) {
+                internal_table.fp_vkDestroyImageView(device, views[j], allocation_callbacks);
+            }
```
th3or14, line 2045: I would call the existing method

```
void Swapchain::destroy_image_views(size_t count, VkImageView const* image_views) {
    for (size_t i = 0; i < count; ++i) {
        internal_table.fp_vkDestroyImageView(device, image_views[i], allocation_callbacks);
    }
}
```

tests/vulkan_mock.hpp
charles-lunarg: Agreed, changing.

th3or14 (approved):

PR #429, opened 2025-09-23
Add Release jobs, at least on Windows they were Debug without specifying CMake --config option

th3or14:
I noticed that in logs of CMake Build step for Windows Release jobs there are folders named Debug. Then I discovered that for MSVC you need to specify `--config Release` in CMake for Release builds. Otherwise, builds are actually Debug even with `CMAKE_BUILD_TYPE=Release`. So I specified `--config <config type>` for Windows jobs.

Then for Release builds I started getting this https://learn.microsoft.com/et-ee/cpp/error-messages/compiler-warnings/compiler-warning-c5072?view=msvc-150 warning. I disabled the warning for Release builds because lack of debug info is not a blocker for using ASAN.

Additionally, specified `--config <config type>` for non-Windows jobs as well. I didn't check if it's necessary for them.


charles-lunarg (approved):
I am very familiar with how MSVC handles debug & release, but I freely admit I didn't think about it when looking at the CI job. I even do the right thing in other github actions in other repos. Thank you for fixing this!

PR #440, opened 2025-11-11
Compile As A Shared Library

StrideStudios:
Added the optional ability to compile VkBootstrap as a shared library

Before this, if the VkInstance was split between two different shared libraries, it would crash since the vulkan_functions singleton would be created twice (once per dll).  This fixes it by exporting it within a shared library.

```
option(VK_BOOTSTRAP_SHARED "Compile Vk Bootstrap as a Shared Library" OFF)
```
Shared Library is Opt in, and otherwise it will compile as static with minimal changes to code.

```
#if !defined(VK_SHARED)
#define VK_API
#else
#if defined(__GNUC__) || defined(__GNUG__) || defined(__clang__)
#define VK_API __attribute__((visibility("default")))
#elif defined(_MSC_VER)
#define VK_API __declspec(dllexport)
#else
#define VK_API
#endif
#endif
```

If Static, the VK_API expands to nothing, so most lines are unaffected.

```
// Export Singleton so it works with shared libraries
VK_API class VulkanFunctions& vulkan_functions();
```

This is the only line different when compiled as a static library, but it functions the same as it did before.

- All Tests pass
- Compiled with MSVC in personal project, and singleton issue was fixed

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,8 +18,6 @@ option(VK_BOOTSTRAP_INSTALL "Enable installing of vk-bootstrap" ${PROJECT_IS_TOP
 
 option(VK_BOOTSTRAP_FORCE_FETCH_CONTENT "Forces Vulkan-Headers to be acquired using Fetch Content")
 
-option(VK_BOOTSTRAP_SHARED "Compile Vk Bootstrap as a Shared Library" OFF)
-
 set(VK_BOOTSTRAP_VULKAN_HEADER_DIR "" CACHE FILEPATH "Specify the location of the Vulkan-Headers include directory.")
 mark_as_advanced(VK_BOOTSTRAP_VULKAN_HEADER_DIR)
 
@@ -98,7 +96,7 @@ add_library(vk-bootstrap::vk-bootstrap ALIAS vk-bootstrap)
 
 # Tell Code that we are a shared library
 if (BUILD_SHARED_LIBS)
-        target_compile_definitions(vk-bootstrap PUBLIC VK_SHARED)
+    target_compile_definitions(vk-bootstrap PUBLIC VKB_BUILD_SHARED)
 endif()
 
 target_include_directories(vk-bootstrap PUBLIC
--- a/src/VkBootstrap.h
+++ b/src/VkBootstrap.h
@@ -71,15 +71,15 @@
 
 // If compiled as a shared library, symbols need to be exported
 
-#if !defined(VK_SHARED)
-#define VK_API
+#if !defined(VKB_BUILD_SHARED)
+#define VKB_EXPORT
 #else
 #if defined(__GNUC__) || defined(__GNUG__) || defined(__clang__)
-#define VK_API __attribute__((visibility("default")))
+#define VKB_EXPORT __attribute__((visibility("default")))
 #elif defined(_MSC_VER)
-#define VK_API __declspec(dllexport)
+#define VKB_EXPORT __declspec(dllexport)
 #else
-#define VK_API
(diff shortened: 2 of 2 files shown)


src/VkBootstrap.h
```diff
@@ -69,6 +69,20 @@
...
 
+// If compiled as a shared library, symbols need to be exported
+
+#if !defined(VK_SHARED)
+#define VK_API
+#else
+#if defined(__GNUC__) || defined(__GNUG__) || defined(__clang__)
+#define VK_API __attribute__((visibility("default")))
+#elif defined(_MSC_VER)
+#define VK_API __declspec(dllexport)
```
charles-lunarg: Can you test that this works with 32 bit Windows? This is important to test because 32 bit Windows ABI prepends underscores on functions (even if you use `extern "C"`) so calling `GetProcAddress("<function_name>");` will fail. But we aren't expecting users to `LoadLibrary` vk-bootstrap, so I don't know if the ABI is problematic.

CMakeLists.txt
```diff
@@ -87,8 +93,16 @@ if(NOT VK_BOOTSTRAP_DISABLE_WARNINGS)
...
+if(VK_BOOTSTRAP_SHARED)
+    add_library(vk-bootstrap SHARED src/VkBootstrap.h src/VkBootstrap.cpp src/VkBootstrapDispatch.h src/VkBootstrapFeatureChain.h src/VkBootstrapFeatureChain.inl)
+    add_library(vk-bootstrap::vk-bootstrap ALIAS vk-bootstrap)
+
+    # Tell Code that we are a shared library
+    target_compile_definitions(vk-bootstrap PUBLIC VK_SHARED)
+else()
+    add_library(vk-bootstrap STATIC src/VkBootstrap.h src/VkBootstrap.cpp src/VkBootstrapDispatch.h src/VkBootstrapFeatureChain.h src/VkBootstrapFeatureChain.inl)
+    add_library(vk-bootstrap::vk-bootstrap ALIAS vk-bootstrap)
+endif()
```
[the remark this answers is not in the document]
StrideStudios: I was unaware you could use add_library without a library type, just testing that and it works fine!

src/VkBootstrap.h
StrideStudios: I'm not sure, I haven't done any testing in 32 bit windows.  I imagine that win32 programs would just have to modify the function names they use ("_foo" instead of "foo").  

A solution would be to construct a .def file, but I don't think it would be a very good option.

src/VkBootstrap.h
```diff
@@ -69,6 +69,20 @@
 #define VKB_VK_API_VERSION_1_0 VKB_MAKE_VK_VERSION(0, 1, 0, 0)
 #endif
 
+// If compiled as a shared library, symbols need to be exported
+
+#if !defined(VK_SHARED)
+#define VK_API
```
charles-lunarg: Going to be that guy and suggest we rename it to VKB_EXPORT to remove confusion of who is defining the macro and prevent possibly collisions.

CMakeLists.txt
```diff
@@ -7,13 +7,19 @@ project(VulkanBootstrap
     DESCRIPTION "A Vulkan utility library to ease the initialization steps in Vulkan"
     VERSION ${VK_BOOTSTRAP_SOURCE_HEADER_VERSION})
 
+# Occasionally Options DO NOT WORK since they don't get updated in the cached cmake files
```
charles-lunarg: Is this comment needed now that the option to build as a shared library is using the BUILD_SHARED_LIBS option?

PR #441, opened 2025-11-13
Remove the use of globals in implementation

charles-lunarg:
Queries and propagates all of the required function pointers into each struct that needs it, removing the need for a global function/variable to store the loaded vulkan functions.

This does require changing the API of the SwapchainBuilder constructor which takes Vulkan handles directly to include the VkInstance, since otherwise it is impossible to load the required functions.

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -139,25 +139,14 @@ std::vector<vkb::detail::FeaturesChain::StructInfo>::const_iterator FeaturesChai
 }
 
 
-class VulkanFunctions {
-    private:
-    std::mutex init_mutex;
-    bool initialized = false;
-
-    std::mutex instance_functions_mutex;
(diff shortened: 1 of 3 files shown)


th3or14:
Shouldn't `VulkanLibrary` have a destructor with calling at least `unload_vulkan_library()` inside?

charles-lunarg:
> Shouldn't VulkanLibrary have a destructor with calling at least unload_vulkan_library() inside?

No because that would break all of the function pointers that have been loaded. Vk-bootstrap never bothered to clean up the libvulkan handle before, and because of the nature of the `vkb::` objects not doing any destruction of vulkan objects, making apps keep their `vkb::` objects alive would be a breaking change. 

I do admit this isn't ideal and leaking dynamic libraries isn't great, but its the corner I've gotten myself into.

Issue #448, opened 2025-12-18
Failing to find extension does not return any useful information


petar-andrejic:
I've been trying to setup a new project using this, and naturally I've been making configuration mistakes which result in my requested extension not being found. However figuring out that this was the reason my instance failed to create was actually somewhat tricky, in that no failure reason is actually returned when an extension is not found.

Specifically, when no extension is found, this function returns false:

https://github.com/charles-lunarg/vk-bootstrap/blob/11ddd832c20960332f4945831e6c985632732c2a/src/VkBootstrap.cpp#L425-L433

This then returns to here, which simply returns an error code without any useful information in the detailed failure reason vector
https://github.com/charles-lunarg/vk-bootstrap/blob/11ddd832c20960332f4945831e6c985632732c2a/src/VkBootstrap.cpp#L775-L778

Preferably, this would set the detailed failure reasons as a list of missing extensions or similar

PR #449, opened 2025-12-18
Return unsupported extension list as detailed failure reason

petar-andrejic:
Suggested fix for #448


charles-lunarg (approved):
This is great! Very much a needed missing feature. The work to get 'missing exts/features' into the physical device creation took significant effort, which led me to forget to apply the same principle equally elsewhere. Thank you for fixing that!

PR #485, opened 2026-06-12
Add Swapchain::get_images_and_image_views helper function

### LINKED ISSUE #484: Add Swapchain::get_images_and_image_views helper function
Would you be open to a combined `Swapchain::get_images_and_image_views`? Like `Device::get_queue_and_index` and `Device::get_dedicated_queue_and_index` added in #463, I feel that asking for both is common enough to warrant the function. It would be convenient to use and reduce redundant work (i.e., `get_image_views` already calls `get_images` internally).

It would also be fairly trivial to implement; the function body would be nearly identical to that of `get_image_views`, with the only differences being the return statement (i.e., `return std::pair<std::vector<VkImage>, std::vector<VkImageView>>(std::move(swapchain_images), std::move(views));`) and removing the `const` qualifier from `swapchain_images_ret` and `swapchain_images` to actually enable the move.

I'd be happy to make the pull request if you approve of this addition!

Hanburgeric:
Adds `Swapchain::get_images_and_image_views()`, which returns both the images and their image views as a `std::pair` for convenience and avoids the redundant call to `get_images()` when both are needed.

Mirrors the existing `get_image_views()` overloads (with and without a pNext chain) and follows the pattern of `Device::get_queue_and_index` from #463.

Closes #484

### DIFF
--- a/src/VkBootstrap.cpp
+++ b/src/VkBootstrap.cpp
@@ -2087,9 +2087,17 @@ Result<std::vector<VkImage>> Swapchain::get_images() {
 }
 Result<std::vector<VkImageView>> Swapchain::get_image_views() { return get_image_views(nullptr); }
 Result<std::vector<VkImageView>> Swapchain::get_image_views(const void* pNext) {
-    const auto swapchain_images_ret = get_images();
+    auto images_and_image_views_ret = get_images_and_image_views(pNext);
+    if (!images_and_image_views_ret) return { images_and_image_views_ret.error() };
+    return std::move(images_and_image_views_ret.value().second);
+}
+Result<std::pair<std::vector<VkImage>, std::vector<VkImageView>>> Swapchain::get_images_and_image_views() {
+    return get_images_and_image_views(nullptr);
+}
+Result<std::pair<std::vector<VkImage>, std::vector<VkImageView>>> Swapchain::get_images_and_image_views(const void* pNext) {
+    auto swapchain_images_ret = get_images();
     if (!swapchain_images_ret) return swapchain_images_ret.error();
-    const auto& swapchain_images = swapchain_images_ret.value();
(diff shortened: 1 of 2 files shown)


src/VkBootstrap.cpp
```diff
@@ -2135,6 +2135,59 @@ Result<std::vector<VkImageView>> Swapchain::get_image_views(const void* pNext) {
...
+        VkResult res = internal_table.fp_vkCreateImageView(device, &createInfo, allocation_callbacks, &views[i]);
+        if (res != VK_SUCCESS) {
+            // Cleanup already created image views
+            destroy_image_views(i, views.data());
+            return Result<std::pair<std::vector<VkImage>, std::vector<VkImageView>>>{ SwapchainError::failed_create_swapchain_image_views,
+                res };
+        }
+    }
+    return std::pair{ std::move(swapchain_images), std::move(views) };
+}
```
charles-lunarg, line 2149: This looks like a straight copy-paste of the `get_image_views()` function, which isn't good. If an issue arises with `get_image_views()`, I would like to be able to fix it in the one place that logic occurs, not copy-paste the fix down to `get_images_and_image_views()`.

charles-lunarg:
Oh, and you'll need to rebase on main to get the fix for windows CI.

src/VkBootstrap.cpp
Hanburgeric: Oh you're right, thank you.

Refactored such that now `get_images_and_image_views()` contains the image view creation logic originally from `get_image_views()`, which now instead delegates to it and returns just the image views.

Also rebased onto main to pick up both the Windows CI and pNext chain fixes.

charles-lunarg (approved):
Much better, thanks!

